PR #37, opened 2017-02-28
`recovr` is not a valid sync strategy

russelldb-bet365:
If you follow the instructions in `Running Leveled` the error:

    {error,{{badmatch,{error,{{{case_clause,recovr},

is shown. `none` seems to work fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ A new database can be started by running
 {ok, Bookie} = leveled_bookie:book_start(RootPath, LedgerCacheSize, JournalSize, SyncStrategy)   
 ```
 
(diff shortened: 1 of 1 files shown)


martinsumner:
D'oh.

Sorry, got confused between two strategy types.  sync_strategy should be either none, riak_sync or sync.

recovr is an option for reload_strategy.

there's more explanation in the leveled_bookie:book_start comments

russelldb-bet365:
>> there's more explanation in the leveled_bookie:book_start comments

yeah, that's how I figured out `none`. Happy to be able to contribute in a small way, thanks for the merge!

PR #50, opened 2017-03-20
Mas sstbinarytrim i42

martinsumner:
This combines two changes, which combined increase throughput under load by around 7%, and also reduce significantly average 2i query response times (under stress).

The first change is to handle the binary blocks that have been read off disk by the SST outside of the SST process -to avoid the copy thta would otherwise follow the conversion to terms.

The second change is to simplify greatly the code when trimming a set of 2i query results, and avoiding the unnecessray opening of some blocks.

With regards to the latter change, the habit of right trimming still seems dubious and may be better handled in penciller when folding.


martinsumner:
There are some hangover improvements here.

One is stopping the whole right trim thing (just check endkey_passed in the fold) and stop the fold when the next key is passed the end_key.

The second is to replace the existing hash index in no-Lookup slots with an index of the first_key in each block - so where the result set in the slot is within a single block, only a single block opening is required.

PR #53, opened 2017-03-20
Switch to binary vclock

martinsumner:
Update pull to include test change


martinsumner:
Previous pull had a commit missing

PR #57, opened 2017-04-06
Mas metadatabin

martinsumner:
Add logic to timeout long-running snapshots.

### DIFF
--- a/include/leveled.hrl
+++ b/include/leveled.hrl
@@ -57,6 +57,7 @@
                         snapshot_query,
                         bookies_mem :: tuple(),
                         source_penciller :: pid(),
+                        snapshot_longrunning = true :: boolean(),
                         levelzero_cointoss = false :: boolean()}).
 
 -record(iclerk_options,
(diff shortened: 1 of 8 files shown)


martinsumner:
This branch mixes two changes (sorry).

The first change is a simple one to reduce the times the code does a binary_to_term on receipt of data, only to ignore that data, and do a term_to_binary when the request is fetched.  So the object metadata is now maintained in leveled as binaries.

Testing this showed improvements, but also exacerbated the issue of growing use of memory when using the mixed 2i test.  The mixed 2i test ended up with 27% of memory being used by the beam, with the majority of it being on-heap binary, and a significant minority of it being process based.

Looked through "Erlang in Anger", and used riak attach-direct - and there were way more processes than expected.  the processes with the most memory weren't the problem, it was the tail of processes  - and there were lots of SST FSM processes that were no longer in the manifest.  Logs showed that some SST files were not being cleared, which pointed at snapshots not being released.

So the next change here was to timeout snapshots that have not been released, and log about the discovery of timed out snapshots,and attempt to release snapshots that did not get a match.  both of these events were occurring, but for less than 1 in a 1000 snapshots.  The two events didn't correlate.

Why these events are still occurring? https://github.com/martinsumner/leveled/issues/58

But for now, the timeout is a workaround.  Now the 2i test ends with the beam having < 4% of memory.

martinsumner:
Accidentally closed ... meant to merge.  What was I doing?

PR #70, opened 2017-05-26
Release penciller snapshot after journal compaction

martinsumner:
As otherwise memory consumption beocmes an issue, as they will take an
hour to timeout naturally.


martinsumner:
This has made a significant improvement in the stability of the memory footprint.

PR #71, opened 2017-05-30
Mas specs i61b

martinsumner:
Add specs and docs


martinsumner:
This gives basic coverage for specs to assist the dialyzer across the external APIs.  Also adds doc comments

PR #77, opened 2017-06-09
Mas compactiontesting

martinsumner:
Add results of volume tests with compaction of the Journal


russelldb:
nice "uptick" at the end of 12hrs for leveled. Wonder if it holds steady thereafter. Longer tests on the horizon?

martinsumner:
Yes.  Want to test with different ring-sizes first though.  Interested in measuring what the penalty is for running a ring-size which is "too big".

PR #78, opened 2017-07-09
Some proposed edits

russelldb:
I couldn't see a good way to comment on the commits themselves as a whole finished piece, so instead I made the edits I _think might work_ and you can keep or ignore as you wish.

Nice article. A+++ would read again :D


martinsumner (approved):
I may re-edit the bit about XOR vs concatenation, but other than that I'm happy with the changes.

PR #80, opened 2017-07-31
Fix dialyzer warnings for newer erlang releases

Licenser:
With recent releases, dialyzers behavior for records changed slightly. While fields used to get an implicit `| undefined` as their type this now needs to be made explicit.

i.e. in R17 and below the following two lines are identical from dialyzers perspective
```erlang
-record(state, {value :: integer()}).
%%% resulted in the type:
-record(state, {value :: integer() | undefined}).
```

in R18 and upwards they are not.

This PR adds the `| undefined` to all records that require them to prevent ~200 warnings when executing dialyzer w/ R19

### DIFF
--- a/include/leveled.hrl
+++ b/include/leveled.hrl
@@ -51,12 +51,12 @@
                         max_run_length}).
 
 -record(penciller_options,
-                        {root_path :: string(),
-                        max_inmemory_tablesize :: integer(),
+                        {root_path :: string() | undefined,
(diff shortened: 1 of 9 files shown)


martinsumner:
Many thanks Heinz.

I've not yet even run leveled on OTP19, been in the to-do stack for a while, so great to have an external contribution.

Licenser:
Glad I could help :)

PR #81, opened 2017-07-31
Add compatibility for old and new random / rand functions

Licenser:
This adds compatibility for both old an d new style random functions and prevents deprecation warnings this way.


martinsumner:
I might be being stupid/tired, but it didn't pass dialyzer for me - and I needed to make some changes to get it to pass https://github.com/martinsumner/leveled/commit/68b2c39cab275d8b0236ad3be889df7656e2dd17

Licenser:
I couldn't run then as I wrote that when #80 wasn't merged, I'll rebase and make sure dialyzer is happy.

Licenser:
Okay I updated the branch but dialyzer doesn't complain here, what's the issue you're seeing?

martinsumner:
I may have done something stupid last night, my git fu is close to zero, but dialyzer complained that uniform_s wasn't used (which it didn't seem to be as it wasn't exported), and then that there was no crpyto:strong_rand/1 function.

Was I testing the right version?  This is what I was looking at:

https://github.com/project-fifo/leveled/blob/38e9b0e80a937ab1775cdb3e0214d5c304a023b5/src/leveled_rand.erl

Licenser:
Oh no you're absolutely right! I dind't see that as I was compiling with the new code :) fixed in a second

martinsumner:
And did you meant crpyto:strong_rand/1 as the OTP18 function for rand_bytes?

```
rand_bytes(Size) ->
    crypto:strong_rand(Size).
```

I couldn't find that function

Licenser:
yes should have been `crypto:rand_bytes` another good catch :)

martinsumner:
Thanks!

PR #105, opened 2017-11-13
Mas otp19 dialyzer

martinsumner:
Since High Sierra upgrade having a separate line for platform_define 17 caused issues.

OTP 19 seemingly matched against it.  OTP 18 had a redefine issue as it appeared to match against it as well.

Using this regex instead seems to work.  Why? Why was this not an issue before High Sierra?

### DIFF
--- a/include/leveled.hrl
+++ b/include/leveled.hrl
@@ -48,8 +48,8 @@
                         source_inker :: pid() | undefined,
                         reload_strategy = [] :: list(),
                         waste_retention_period :: integer() | undefined,
-                        compression_method :: lz4|native,
-                        compress_on_receipt :: boolean(),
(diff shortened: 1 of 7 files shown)


martinsumner:
Primarily this pull request resolves a series of dialyzer errors in OTP 19.  Looks like previously I'd not run dialyzer on all versions (18.3 dialyzer showed no issues) so missed this.

There is also a hack for an issue whereby the platform_define statements stopped working in High sierra (but not other Mac OSX versions)!  this could be fixed by removing the line for matching on OTP release 17 (as OTP 19 and 18 both seemed to match on this line).  for now  anew regex line has been added to platform_define - as this appears to work across OSX versions.  The underlying mystery remains unresolved.

PR #109, opened 2017-11-20
Mas i107 foldobjects2d

martinsumner:
sometimes ct tests don’t hit this - surprisingly


martinsumner:
This PR allows for object folds to be requested in SQN order.  This means that object folds can be done across the journal (bitcask style), rather than across the ledger (leveldb style).

It is assumed, that if fadvise is followed correctly - SQN order would be quicker for allkeys folds.  But this is yet to be proven.

PR #113, opened 2017-11-22
Mas i111 ssttimings

martinsumner:
Now all timing points should be made in a consistent fashion


martinsumner:
Te use of periodic sampling introduced in leveled_cdb seemed a more effective way of tracking performance pain points through logging - so this approach is now used elsewhere.

PR #116, opened 2018-02-06
remove warnings_as_errors from erl_opts - would like to compile level…

bryanhuntesl:
…ed despite warnings

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,5 +1,4 @@
-{erl_opts, [warnings_as_errors,
-            {platform_define, "^1[7-8]{1}", old_rand},
+{erl_opts, [{platform_define, "^1[7-8]{1}", old_rand},
             {platform_define, "^R", old_rand},
             {platform_define, "^R", no_sync}]}.
 


martinsumner:
Oh, sorry.  I thought it did.  What OTP version you trying? I think it works on 16 - 19.

bryanhuntesl:
I'm going all out crazy and trying to run it one 20 :-D

martinsumner:
Ah yes.  I've never tried that.  Raise an issue if you want and I will try and look at it.  

OTP18.3 is the version I originally built it on, so that's normally a safe bet.  Heinz then ported to 19, and I back-ported to 16 for Riak support.

You happy me to ignore the PR in the mean time, as I'd like to keep breaking it on warnings for the supported versions for now?

bryanhuntesl:
I’m trying to include it in an Elixir application - they don’t have the means to override rebar erl_opts as far as I can determine - that’s why I put the PR - otherwise I would need to work off of a fork.

martinsumner:
Fork away for now.  It is still not  at 1.0, so I don't guarantee I won't put breaking changes into the API - so you're probably best working with a fork I guess.

I will look at the OTP 20 issues though.

bryanhuntesl:
Thanks @martinsumner - actually it compiles fine on 20 - with the usual "`gen_fsm` will disappear at some unspecified point in the future" deprecation warnings - I wish they hadn't made that decision. I've started an instance and successfully written/read.

Issue #122, opened 2018-03-15
Periodic failing test leveled_sst:indexed_list_mixedkeys_bitflip_test/0


russelldb:
```Failures:

  1) leveled_sst:indexed_list_mixedkeys_bitflip_test/0
     Failure/Error: {error,
                        {badmatch,
                            <<0,0,1,79,0,0,1,76,0,0,2,213,0,0,1,150,0,0,0,0,
                              59,140,84,131,162,160,253,130,174,188,53,201,
                              234,135,231,234,110,129,106,204,111,222,232,207,
                              109,139,62,139,31,175,241,239,181,235,45,232,67,
                              248,217,143,157,224,110,178,233,165,229,158,176,
                              198,240,212,94,175,109,138,123,254,18,225,178,
                              136,235,195,81,221,193,131,80,0,0,8,15,120,1,
                              173,213,185,78,195,64,16,6,224,177,179,118,14,
                              168,184,4,66,226,172,160,64,89,95,68,17,5,162,
                              69,188,2,145,137,141,188,216,178,41,236,72,169,
                              169,169,40,168,104,225,9,104,233,121,34,90,102,
                              52,91,35,138,233,182,250,52,251,207,111,111,5,0,
                              219,133,91,168,12,28,83,130,127,213,205,203,188,
                              45,92,60,182,122,102,234,54,237,149,160,174,243,
                              101,92,168,212,43,220,12,252,116,222,154,69,158,
                              193,192,212,247,166,54,237,50,131,97,221,204,
                              170,166,41,187,199,12,84,221,85,213,95,226,29,
                              128,163,25,77,16,245,197,208,105,9,30,78,26,5,
                              168,158,138,169,55,172,134,33,170,135,98,234,11,
                              171,113,132,106,34,165,186,1,171,122,130,234,
                              154,148,218,59,99,53,160,10,236,136,169,207,54,
                              87,74,224,72,74,85,159,118,86,234,192,150,148,
                              234,221,114,93,207,17,237,75,161,254,37,163,26,
                              81,71,10,237,43,123,127,82,55,197,212,87,86,147,
                              49,170,23,82,234,96,221,54,128,118,117,32,166,
                              30,219,4,104,89,187,98,234,19,171,17,125,89,99,
                              49,245,141,213,152,254,131,19,41,117,248,206,
                              170,166,109,141,196,212,47,155,0,53,235,68,76,
                              253,182,9,208,172,129,148,58,250,176,42,109,107,
[...]

PR #123, opened 2018-03-15
Fix failing test, I don't understand how

russelldb:
See #122 for details

### DIFF
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -1519,7 +1519,7 @@ crc_check_slot(FullBin) ->
         PosBL:32/integer, 
         CRC32H:32/integer,
         Rest/binary>> = FullBin,
-    PosBL0 = min(PosBL, byte_size(FullBin) - 3), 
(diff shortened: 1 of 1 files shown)


martinsumner:
Just to be clear on the how.  The purpose of crc_check_slot/1 is to check that the binary on the disk is a a slot which has not been corrupted.  The slot is made up of a header, and some blocks, and the blocks are separately CRC checked, so just the header needs to be CRC checked here

At the head of a slot are 3 4-byte integers for giving the CRC of the header, the byte position of the header, and a CRC check of the byte position of the header.  The code extracts the header before checking the byte position used to extract the header (for reasons which aren't clear - perhaps to make the test coverage easier to complete), and so has to handle the position being wrong and bigger than the binary from which it is trying to extract the header.

The line incorrectly set the maximum size of the position to be 3 bytes less than the binary length, not 3 x 4 bytes less than the binary length.

This PR corrects this, but on reflection it exposes that the function was mis-written in the first place - the CRC check of the position should be made before the header is extracted (rendering the minimum check unnecessary), and this should be tested consistently each test run.

Issue #31, opened 2017-02-28
CRC in CDB Files


martinsumner:
The constant database files are genuine constant database files, as the CRC check is wrapped in to the CDB file.  This means that the journal file could not be interrogated by another CDB library.

This is unnecessary.  The CRC is at the wrong layer of abstraction and should be added prior to pushing by the Inker, and when scanning the Inker should provide a validate function to the CDB process that does the CRC check.

martinsumner:
The object compression using binary_to_term uses zlib ... but doesn't necessarily checksum.  You can bit-flip  in a binary, and get the bit-flipped result even if it was created with compression.

I speculate that if the term cannot be compressed, it isn't compressed and so there will be no CRC check. 

A CRC check is necessary, regardless of the the use of term_to_binary compression.

martinsumner:
Also the Key is not CRC checked.  It may crash in an unexpected place (for example if it passes binary_to_term, but is a tuple with the wrong number of elements)

martinsumner:
https://github.com/martinsumner/leveled/pull/140

This doesn't resolve the compatibility problem, in that another CDB reader would have to discard/ignore the checksum in the application.  However, cdb compatibility is not a core goal.  

This PR stops the crashes on corrupted keys, and so the issue is closed.

PR #135, opened 2018-05-04
Mas i61 codectidy

martinsumner:
Declare and export shared types within leveed_codec, and popular more functions with -spec definitions.


martinsumner:
note that [this](https://github.com/martinsumner/leveled/issues/31) still needs to be examined - especially CRC for keys in leveled_cdb.  There was some vague/inadequate mitigation for this in the code prior to this PR, which is now gone.

PR #137, opened 2018-05-16
Mas i11 backenddelete

martinsumner:
An issue was spotted.  If we use a segment filter in a query, and there are multiple matches within a given slot - only the first match is returned.

Tests didn't detect this.  Now they do, and the issue is resolved.


martinsumner:
Some further work needs to be done to consider the scenario when the block index is not cached.  What if it never gets cached (e.g. when leveled is used as a key-ordered backend for AAE)

PR #150, opened 2018-07-11
Mas i29 linkall

martinsumner:
If a snapshot breaks a penciller clone, this shouldn't crash the main process.


martinsumner:
If an individual process in leveled dies, then the whole tree of processes should die.  Responsibility lies up the tree from book_start to restart.

For snapshots these are not linked to the original tree.  as snapshots should be forgotten by the tree on failure by timeout and should not crash the main process.

PR #167, opened 2018-09-06
Provide a top level API for folds

russelldb:
As the fold functions have been added to get_runner in an ad hoc way,
naturally, given the ongoing development of levelEd to support Riak,
it was difficult for a new user (in this case Quviq) to see what folds
are supported, and with what arguments, and expectations.

This PR is for discussion. It is one of many ways to group, spec, and
document the fold functions.

A test is also added for coverage of range queries.

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -58,7 +58,6 @@
         book_get/4,
         book_head/3,
         book_head/4,
-        book_returnfolder/2,
         book_snapshot/4,
         book_compactjournal/2,
         book_islastcompactionpending/1,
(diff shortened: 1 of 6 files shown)


martinsumner:
I'm happy with the @doc statements.  Couple of typos, but I think they can fixed later.

Is the intention to keep book_returnfolder/2 as-is - so thinks like riak_kv_leveled_backend continue to work as before?  Does this need to be marked as deprecated, to remind us to migrate the kv_backend code?

russelldb:
It's a good question. The thing about `book_returnfolder` is  that it is helpful for the time being, if you want to add new folds and don't know how or where they fit into the API. For a project under rapid development it's valuable to have a super flexible (less strictly defined) interface. I wouldn't deprecate until 1.0

PR #168, opened 2018-09-09
Mas i164 hotbackup

martinsumner:
Add hot backup capability.

This is to make sure that backups can be taken safely/consistently.


martinsumner:
based on [mv-hot-backup](https://github.com/basho/leveldb/wiki/mv-hot-backup)

PR #172, opened 2018-09-17
Mas iq6 rootpathundefined

martinsumner:


martinsumner:
https://github.com/Quviq/leveled/issues/6

If you start a Bookie with an undefined root path - it should no shutdown with {error, no_root_path} - ee leveled_bookie:undefined_rootpath_test/0

PR #173, opened 2018-09-18
Callers of bucket list expect the traversal to be in order

russelldb:
Due to the internal fold over buckets returning an un-reversed
accumulator, the API bucketlist code caller's fold fun traversed the
bucket list in reverse order. This lead to some inconsistencies when
comparing a buckelist of all buckets, vs, first bucket only. i.e. the
'first' bucket passed to the foldfun was in fact the last bucket read
from the ledger.


russelldb:
See https://github.com/Quviq/leveled/issues/9

PR #214, opened 2018-12-04
Add skeleton _app and _sup modules

uwiger:
Given that leveled has a `mod` attribute, and thus should be runnable as an OTP application, a few modules were missing.


uwiger:
Addressed in #216

PR #218, opened 2018-12-05
Mas uw configurableloglevel

martinsumner:
Based on https://github.com/martinsumner/leveled/pull/215

This allows for log level to be set at startup, as well as for the log level to be set through set_env.  Also allows for forced_logs - a list of log references that will always be logged even when the log level is set to a higher level.  Currently stats are not output other than via logs - so setting log_level to error would deny access to stats, by having the stats logs as forced_logs.

The default log level remains at info.  Other may differ in opinion as to what the appropriate default level is - but the ability to start with another level means that all opinions can be supported.

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -311,12 +311,12 @@
             % As an alternative a higher log_level can be used to reduce this
             % 'noise', however, there is currently no separate stats facility
             % to gather relevant information outside of info level logs.  So
-            % moving to higher log levels will at present make the opertaor
+            % moving to higher log levels will at present make the operator
             % blind to sample performance statistics of leveled sub-components
             % etc
         {forced_logs, list(string())}
             % Forced logs allow for specific info level logs, such as those
-            % logging stats to be logged even when the defualt log level has
+            % logging stats to be logged even when the default log level has
             % been set to a higher log level.  Using:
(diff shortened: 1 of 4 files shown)


src/leveled_bookie.erl
```diff
@@ -293,12 +296,33 @@
...
+        {log_level, debug|info|warn|error|critical} |
+            % Set the log level.  The default log_level of info is noisy - the
+            % current implementation was targetted at environments that have
+            % facilities to index large proportions of logs and allow for
+            % dynamic querying of those indexes to output relevant stats.
+            %
+            % As an alternative a higher log_level can be used to reduce this
+            % 'noise', however, there is currently no separate stats facility
+            % to gather relevant information outside of info level logs.  So
+            % moving to higher log levels will at present make the opertaor
```
uwiger: operator

src/leveled_bookie.erl
```diff
@@ -293,12 +296,33 @@
...
+            % blind to sample performance statistics of leveled sub-components
+            % etc
+        {forced_logs, list(string())}
+            % Forced logs allow for specific info level logs, such as those
+            % logging stats to be logged even when the defualt log level has
```
uwiger: default

uwiger (approved):

Issue #223, opened 2018-12-07
rebar3 eunit fails on OTP 21.1


uwiger:
Some sample output:

```
===> Performing EUnit tests...
.........=CRASH REPORT==== 7-Dec-2018::09:00:29.958959 ===
  crasher:
    initial call: leveled_bookie:init/1
    pid: <0.959.0>
    registered_name: []
    exception exit: no_root_path
      in function  gen_server:init_it/6 (gen_server.erl, line 358)
    ancestors: [<0.660.0>]
    message_queue_len: 0
    messages: []
    links: []
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 233
    stack_size: 27
    reductions: 233
  neighbours:
```
and a bit later:

```
.........................F=ERROR REPORT==== 7-Dec-2018::09:01:38.280915 ===
** Generic server <0.1234.0> terminating 
** Last message in was {'$gen_cast',{release_snapshot,<0.1239.0>}}
** When Server state == {state,
                            [{1,
                              [{3,"../test/journal/journal_files/nursery_3",
                                <0.1237.0>,
                                {4,stnd,{o,<<"B">>,"Key4",null}}},
                               {1,"../test/journal/journal_files/nursery_1",
                                <0.1236.0>,
                                {2,stnd,{o,<<"B">>,"Key2",null}}}]}],
                            1,4,<0.1237.0>,[],
                            [{<0.1239.0>,1}],
                            "../test/journal",
                            {cdb_options,300000,undefined,undefined,true,sync},
                            <0.1235.0>,false,undefined,false,native,true,
                            undefined}
** Reason for termination == 
** {terminated,[{io,format,
                    [<0.1230.0>,
                     "2018-12-07T09:01:38.280 I0004 ~w Remaining number of journal snapshots is ~w~n",
                     [<0.1234.0>,0]],
                    []},
                {leveled_inker,handle_cast,2,
                               [{file,"/home/uwiger/tmp/test/leveled/src/leveled_inker.erl"},
                                {line,698}]},
                {gen_server,try_dispatch,4,
                            [{file,"gen_server.erl"},{line,637}]},
                {gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,711}]},
                {proc_lib,init_p_do_apply,3,
                          [{file,"proc_lib.erl"},{line,249}]}]}
```

[...]

PR #224, opened 2018-12-07
warning-free vsns of string:str/2 & string:right/3

uwiger:
It seems as if OTP is in the process of deprecating `string:str/2` and `string:right/3` (building OTP from the latest sources will give warnings, which results in build failure due to `warnings_as_errors`, though OTP 21.1 does not give these warnings.)

These changes offer compatible implementations of `string:str/2` and `string:right/3`. Perhaps unnecessary to conditionally compile ... Without conditional compilation, the changes pass eunit tests on OTP 20.3. On 21.1 there are other failures unrelated to this PR (also on master, see issue #223).


martinsumner:
I'm closing this as there are alternative PRs that cover this work:

https://github.com/martinsumner/leveled/pull/232
https://github.com/martinsumner/leveled/pull/231

uwiger:
Ok

PR #231, opened 2018-12-10
Uw avoid set env

uwiger:
Using `application:set_env/3` to store the log level is problematic in two ways:

* If multiple leveled instances are used in the same system, each subsequent log level setting will affect all running instances.
* Whereas the application_controller is mostly non-blocking, it isn't for `application:stop/1`, so any attempt to alter the log level after application:stop/1 is called (e.g. on mnesia) will cause a timeout exception.

This PR instead stores the log settings in the process dictionary of `leveled_bookie`. The processes spawned thereafter read the log settings before spawning and pass them on. A way to dynamically alter the log level is missing.

The change has been tested on an OTP 21 system, where both eunit and ct tests passed.

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -1061,9 +1061,8 @@ init([Opts]) ->
             {InkerOpts, PencillerOpts} = set_options(Opts),
 
             LogLevel = proplists:get_value(log_level, Opts),
-            ok = application:set_env(leveled, log_level, LogLevel),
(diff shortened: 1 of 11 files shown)


martinsumner:
@uwiger I'm going to look at this tomorrow afternoon, but this does look to be a better approach.  Bear with me.

martinsumner:
I have an adjusted version of this PR.  Have a look:

https://github.com/martinsumner/leveled/pull/232

PR #308, opened 2020-03-09
Mas i306 reviseretain

martinsumner:
This is a PR to prepare for the introduction of alternative journal_compact types - in particular one which will safely erase the history of deleted objects.

The building blocks here are:

- The penciller will now return current|replaced|missing rather than true|false when checking if a sequence number is up to date.

- Extra testing has been added that combines alterations, deletions, compactions and starting with empty ledgers.

- The tech debt with regards to the log noise on starting with an empty ledger has been resolved.


ThomasArts:
see comments in PR #310.

PR #310, opened 2020-03-24
Mas i306 implementrecalc

martinsumner:
The `recalc` strategy for reloading the ledger from the Journal, was previously referenced but not implemented or tested.

This PR adds an implementation of the strategy for the ?RIAK_TAG, with appropriate test coverage.

Before adding an object to the leveled backend, the application is expected to fetch the HEAD of the previous version of the object and include IndexSpecs alongside the PUT - a list of [{add, Field, Term}, {remove, Field, Term}] entries indicating which indexes should be added or removed as part of the change.

These key changes are included in the Journal entry for the change.  When the Journal is compacted,  and the `retain` compaction strategy is used, then those key changes are kept even when the object is replaced (or deleted).  So the entry is slimmed from the full object, to a skeleton object just containing the index key changes.  So the Journal will eventually contain the most recent object, plus a history of index changes.  

On startup, the ledger needs to be completed (normally due to the dropping of the ledger cache, but in some circumstances the ledger may be deleted due to corruption or backup recovery).  It does this by re-loading the data in sequence (from the highest sequence number in the Ledger) from the Journal.

The `retain` strategy would load any index changes as well as any object to rebuild the ledger of its previous state.  

The `recalc` strategy now throws away the Index Changes from a replaced object during compaction - and keeps no history.  On startup it will re-calculate the index specs for each object in the Journal, by comparing the journal object metadata with any Ledger object metadata for that key.  This will also bring the Journal and the Ledger into sync.  With `recalc` the reload commits to re-performing the original work of the application to make the Index Specs, rather than relying on the retained Specs.

The `retain` strategy was implemented as a default as it was simpler, and isolated the database from the application logic required to recalculate the index specs.   However, it can lead to lots of "garbage" in the Journal - a legacy of skeleton KeyChanges objects, even after the objects have been deleted.   This "garbage" may significantly increase the cost of scoring files for Journal compaction - and can therefore impact the performance of the store. 

The introduction of UserDefined functions via `leveled_head` has now provided a better answer to the abstraction problem - and `recalc` now allows for safety without the overhead of a full history of skeleton objects.  In some cases startup times may be impacted (in some cases though startup times may actually be faster with `recalc`)


martinsumner:
It should be noted that it is possible to switch a store from the "retain" strategy to the "recalc" strategy by switching the configuration.  However, it is not possible to reverse this change safely.

docs/DESIGN.md
```diff
@@ -120,7 +120,7 @@ Three potential recovery strategies are supported to provide some flexibility fo
 
 - retain - on compaction KeyDeltas are retained in the Journal, only values are removed.
 
-- recalc (not yet implemented) - the compaction rules assume that on recovery the key changes will be recalculated by comparing the change with the current database state. In recovery the key changes will be recalculated by comparing the change with the current database state.
+- recalc - the compaction rules assume that on recovery the key changes will be recalculated by comparing the change with the current database state. In recovery the key changes will be recalculated by comparing the change with the current database state.  A user-defined function should be passed in at startup to achieve this recalculation (to override `leveled_head:diff_indexspeacs/3`).
```
ThomasArts:
```suggestion
- recalc - the compaction rules assume that on recovery the key changes will be recalculated by comparing the change with the current database state. In recovery the key changes will be recalculated by comparing the change with the current database state.  A user-defined function should be passed in at startup to achieve this recalculation (to override `leveled_head:diff_indexspecs/3`).
```

docs/STARTUP_OPTIONS.md
```diff
@@ -77,11 +77,11 @@ However, what if the Ledger had been erased?  This could happen due to some corr
 
 The are three potential strategies:
 
- - `skip` - don't worry about this scenario, require the Ledger to be backed up;
+ - `recovr` - don't worry about this scenario, require the Ledger to be backed up;
```
ThomasArts: There are traces of "skip" strategy in the code:
```
src/leveled_codec.erl:        retain|skip|recalc.
src/leveled_codec.erl:                                                    -> skip|retain|recalc.
src/leveled_iclerk.erl:%% If the strategy is skip, we don't care about KeyDeltas.  Note though, that
src/leveled_iclerk.erl:%% a `skip` strategy.
```

src/leveled.app.src
```diff
@@ -1,7 +1,7 @@
 {application, leveled,
  [
   {description, "Key Value store based on LSM-Tree and designed for larger values"},
-  {vsn, "0.9.20"},
+  {vsn, "0.9.21"},
```
ThomasArts: Since you removed the "skip" strategy, one could wonder if this is a ptach or a minor update. I would go for 0.10.0... but that's fully up to you.

src/leveled_codec.erl
```diff
@@ -358,24 +363,24 @@ endkey_passed(EndKey, CheckingKey) ->
...
-                    ReloadStrategy0,
-                    AltList).
+    lists:ukeymerge(1,
+                    lists:ukeysort(1, AltList),
+                    lists:ukeysort(1, DefaultList)).
 
 
--spec get_tagstrategy(ledger_key(), compaction_strategy()) 
+-spec get_tagstrategy(ledger_key()|tag()|dummy, compaction_strategy()) 
                                                     -> skip|retain|recalc.
```
ThomasArts:
```suggestion
                                                    -> recovr|retain|recalc.
```

ThomasArts: weird that dialyzer did not found it (at least not on my machine)

src/leveled.app.src
martinsumner: It was more of a confusion over whether the strategy was called 'skip' or 'recovr', I think.

martinsumner:
Volume test showed that the difference between running the test with a retain and recalc strategy was well within the margin of error (over the course of a 48 hour test the recalc strategy achieved << 1% extra throughput than the retain strategy).

A reload of a vnode was attempted from an emptied ledger - this was a 43GB vnode.

- with the retain strategy this took 61 minutes.

- with the recalc strategy this tool 45 minutes.

It had been assumed that a negative side effect of the strategy change would be increased startup time in this case - so this is a very encouraging result.

The other noticeable change is that the recalc strategy had a higher proportion of journal compaction jobs at the max run length, and overall spent 19% less time running these jobs.  This is also a positive result.

V&P evidence supports this PR.

Issue #311, opened 2020-03-25
Slow queries when index entries deleted


martinsumner:
If there is a large leveled store, then we add a lot of index entries, then delete those index entries - then the time taken to conclude an index query will be proportional to the number of removed entries not the active entries.

This is because leveled tombstones in the ledger are maintained as the ledger is compacted until the tombstone reaches the basement.  When it reaches the basement, the tombstone will be reaped - as it is now clear that there are no further objects to be erased.

So if we add a {Key1, V1}
then alter by adding {Key1, V2}
then delete with {Key1, tomb}

The tomb must be maintained in the store even after it has merged through {Key1, V2} so as to not resurrect {Key1, V1}.  Only when a tomb is merged into the basement level do we know it has no more victims to kill and so the tomb can be erased i.e. the tomb will not be written to the basement level.

There may be certain workloads that result in very infrequent merges to the basement, and hence there may be large files full of tombstones that are repeatedly scanned for queries to discover all the tombstones (which don't result in any results).  So the result may take a long time to return even when there are very few actual results.

martinsumner:
Within basho leveldb, this situation could also arise, but is more than likely resolved via grooming compactions.  Proactively pushing files with large numbers of tombstones towards the basement.

https://github.com/basho/leveldb/wiki/Mv-aggressive-delete

martinsumner:
There may be an additional problem related to the switching of files.  When a file in Level Basement + 1 is picked to be merged, if it doesn't overlap with any file in Level Basement, then it will be switched to the Basement rather than undergoing a merge.

This switching process would mean that tombstones are not reaped (unless a file in the future is merged from Basement + 1 which does overlap).

martinsumner:
There are a number of alternatives here:

1. Treat index keys differently, as they don't have modifications - and drop them on the first merge with an older key (as they have no value).

2. Force a proportion of compactions to be grooming compactions (ones that favour the SST file with the most tombstones).

3. Add additional grooming compactions (potentially triggered only when the clerk has had no work for a period).

Option 1 may encounter issues with TTL and index entries.  The TTL is the value, so potentially you could have updated index entries where the TTL changes.  This is not part of Riak implementation, but exists as a theoretical possibility.

Option 2 will makes things better without necessarily solving the problem.

[...]

PR #312, opened 2020-03-27
Mas i311 mergeselector

martinsumner:
In response to https://github.com/martinsumner/leveled/issues/311

This PR alters the selection of merge files in the LSM tree.  Previously when Level n in the tree had over-grown its maximum size, a file would be selected at random from the Level to be merged into the Level (n + 1) below.

This strategy has now changed.  Half the time it will pick purely at random.  The other half of the time it will pick up to 8 files at random, and then select the file out of those 8 which has the highest count of tombstones.

To facilitate this SST files will now count their tombstones during creation, and persist this count to the store metadata.  A L0 file doesn't do, this, and a file that had been written in the previous version.  When requested to return their tombstone count, such legacy files will return "not_counted".

A "not_counted" response will be consider superior to an actual response of the count.  The priority after migration is to roll forward to new version files, the comparison between counts will only be made once all files in the eight selected are returning a count.

### DIFF
--- a/src/leveled_pclerk.erl
+++ b/src/leveled_pclerk.erl
@@ -183,7 +183,7 @@ merge(SrcLevel, Manifest, RootPath, OptsSST) ->
             leveled_log:log("PC023",
                             [SrcLevel + 1, FCnt, AvgMem, MaxFN, MaxP, MaxMem])
     end,
-    Src = leveled_pmanifest:mergefile_selector(Manifest, SrcLevel),
+    Src = leveled_pmanifest:mergefile_selector(Manifest, SrcLevel, random),
     NewSQN = leveled_pmanifest:get_manifest_sqn(Manifest) + 1,
     SinkList = leveled_pmanifest:merge_lookup(Manifest,
                                                 SrcLevel + 1,
--- a/src/leveled_pclerk.erl
+++ b/src/leveled_pclerk.erl
@@ -260,9 +260,9 @@ do_merge(KL1, KL2, SinkLevel, SinkB, RP, NewSQN, MaxSQN, OptsSST, Additions) ->
                                                 length(Additions)),
     leveled_log:log("PC012", [NewSQN, FileName, SinkB]),
     TS1 = os:timestamp(),
-    case leveled_sst:sst_new(RP, FileName,
-                                KL1, KL2, SinkB, SinkLevel, MaxSQN, 
-                                OptsSST) of
+    case leveled_sst:sst_newmerge(RP, FileName,
+                                    KL1, KL2, SinkB, SinkLevel, MaxSQN,
+                                    OptsSST) of
         empty ->
             leveled_log:log("PC013", [FileName]),
             do_merge([], [],
--- a/src/leveled_pclerk.erl
+++ b/src/leveled_pclerk.erl
@@ -49,12 +49,15 @@
 
 -define(MAX_TIMEOUT, 2000).
 -define(MIN_TIMEOUT, 200).
+-define(GROOMING_PERC, 50).
 
 -record(state, {owner :: pid() | undefined,
                 root_path :: string() | undefined,
                 pending_deletions = dict:new(), % OTP 16 does not like type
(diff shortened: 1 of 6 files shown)


src/leveled_pclerk.erl
```diff
@@ -285,6 +296,18 @@ do_merge(KL1, KL2, SinkLevel, SinkB, RP, NewSQN, MaxSQN, OptsSST, Additions) ->
...
 
+-spec grooming_scorer(list(manifest_entry())) -> manifest_entry().
+grooming_scorer(Sample) ->
+    ScoringFun = 
+        fun(ME) ->
+            TombCount = leveled_sst:sst_gettombcount(ME#manifest_entry.owner),
+            {TombCount, ME}
+        end,
+    ScoredSample =
+        lists:reverse(lists:ukeysort(1, lists:map(ScoringFun, Sample))),
```
ThomasArts: Not sure whether lists:reverse and lists:ukeysort (traversing the list in N and  N * log(N) are expensive, but you could just do a fold, I guess.
```
grooming_scorer([ME  | MEs]) ->
    {Highest, BestME} =
         grooming_scorer(leveled_sst:sst_gettombcount(ME#manifest_entry.owner), ME, MEs),
         leveled_log:log("PC024", [HighestTC]),
         BestME.

grooming_scorer(HighestTC, BestME, []) ->
     {HighestTC, BestME};
grooming_scorer(HighestTC, BestME, [ME | MEs]) ->
    TombCount = leveled_sst:sst_gettombcount(ME#manifest_entry.owner),
    case TombCount > HighestTC of
        true -> grooming_scorer(TombCount, ME, MEs);
        false -> grooming_scorer(HighestTC, BestME, [ME | MEs])
   end.
```
linear in sample size.

src/leveled_pclerk.erl
```diff
@@ -183,7 +186,15 @@ merge(SrcLevel, Manifest, RootPath, OptsSST) ->
             leveled_log:log("PC023",
                             [SrcLevel + 1, FCnt, AvgMem, MaxFN, MaxP, MaxMem])
     end,
-    Src = leveled_pmanifest:mergefile_selector(Manifest, SrcLevel),
+    SelectMethod =
+        case leveled_rand:uniform(100) of
+            R when R < ?GROOMING_PERC ->
```
ThomasArts:
```suggestion
            R when R =< ?GROOMING_PERC ->
```
I guess that if you happen to set ?GROOMING_PERC to 100, you want to do this always, whereas 0 means, never. Since rand returns number between 1 and 100, a `less equal` is more accurate

src/leveled_pmanifest.erl
```diff
@@ -441,14 +445,28 @@ merge_lookup(Manifest, LevelIdx, StartKey, EndKey) ->
...
+                                            Manifest#manifest.levels)),
+    SelectorFun =
+        fun(_I, Acc) ->
+            {_SK, ME} = lists:nth(leveled_rand:uniform(length(Level)), Level),
+            [ME|Acc]
+        end,
+    Sample =
+        lists:usort(lists:foldl(SelectorFun, [], lists:seq(1, ?GROOM_SAMPLE))),
+    % Note that Entries may be less than GROOM_SAMPLE, if same one picked
+    % multiple times
```
ThomasArts:
```suggestion
    % multiple times
    % Also note that Sample is never empty
```

ThomasArts (approved):

src/leveled_pclerk.erl
martinsumner: OK, thanks, gone with a corrected version of this

src/leveled_pclerk.erl
martinsumner: Agreed and updated

src/leveled_pmanifest.erl
martinsumner: Updated

martinsumner:
Volume test passed, with a small improvement in throughput (though within the margin of error)

PR #320, opened 2020-11-27
Allow for caching of compaction scores

martinsumner:
Potentially reduce the overheads of scoring each file on every run.

The change also alters the default thresholds for compaction to favour longer runs (which will tend towards greater storage efficiency).


priv/leveled.schema
```diff
@@ -143,7 +149,7 @@
 %% then it is a candidate (e.g. in default case if 50% of space would be
```
ThomasArts:
```suggestion
%% then it is a candidate (e.g. in default case if 30% of space would be
```

src/leveled_cdb.erl
```diff
@@ -424,6 +432,20 @@ cdb_isrolling(Pid) ->
 cdb_clerkcomplete(Pid) ->
     gen_fsm:send_all_state_event(Pid, clerk_complete).
 
+-spec cdb_getcachedscore(pid(), erlang:timestamp()) -> undefined|float().
+%% @doc
+%% Return the cached score for a CDB file
+cdb_getcachedscore(Pid, Now) ->
```
ThomasArts, line 438: why not computing the timestamp inside the call? A bit weird to have it passed in through the interface, but probably that's for testing...

```
cdb_getcachedscore(Pid) ->
    gen_fsm:sync_send_all_state_event(Pid, {get_cachedscore, erlang:timestamp()}, infinity).
```

src/leveled_cdb.erl
```diff
@@ -2396,6 +2435,16 @@ get_keys_byposition_manykeys_test_to() ->
     SampleList3 = cdb_getpositions(P2, KeyCount + 1),
     ?assertMatch(KeyCount, length(SampleList3)),
     
+    ?assertMatch(undefined, cdb_getcachedscore(P2, os:timestamp())),
+    ok = cdb_putcachedscore(P2, 80.0),
+    ?assertMatch(80.0, cdb_getcachedscore(P2, os:timestamp())),
+    timer:sleep(1000),
+    {NowMega, NowSecs, _} = Now = os:timestamp(),
+    ?assertMatch(80.0, cdb_getcachedscore(P2, Now)),
```
ThomasArts, line 2443: here you can fake "os:timestamp" by remembering the old and add 1000 to its seconds.
No need to sleep for a second and make testing slower.
Or, what else is there purpose of the API with a timestamp?

src/leveled_iclerk.erl
```diff
@@ -549,44 +584,31 @@ safely_log_filescore(PositionList, FN, Score, SW) ->
     leveled_log:log_timer("IC004", [Score, AvgJump, FN], SW).
 
 -spec size_comparison_score(list(key_size() | corrupted_test_key_size()),
```
ThomasArts, line 586: It's a bit confusing that there is a `leveled_inker:filterfun()` type and a `leveled_cdb:filter_fun()`. Sooner or later this is going to be messed up.

src/leveled_iclerk.erl
```diff
@@ -549,44 +584,31 @@ safely_log_filescore(PositionList, FN, Score, SW) ->
...
-                                {current, _} ->
-                                    {ActSize + Size - ?CRC_SIZE, RplSize};
-                                {_, true} ->
-                                    {ActSize + Size - ?CRC_SIZE, RplSize};
-                                _ ->
-                                    {ActSize, RplSize + Size - ?CRC_SIZE}
-                            end
+                            {ActSize + Size - ?CRC_SIZE, RplSize};
+                        convert ->
+                            {ActSize, RplSize  + Size - ?CRC_SIZE};
```
ThomasArts:
```suggestion
                            {ActSize, RplSize + Size - ?CRC_SIZE};
```

convert and false are actually the same...

src/leveled_iclerk.erl
```diff
@@ -549,44 +584,31 @@ safely_log_filescore(PositionList, FN, Score, SW) ->
...
-                            % there is support for changing strategy from
-                            % retain to recalc
-                            case TS of
-                                retain ->
-                                    {ActSize + Size - ?CRC_SIZE, RplSize};
-                                _ ->
-                                    {ActSize, RplSize + Size - ?CRC_SIZE}
-                            end;
+                    ToRetain =
+                        to_retain({SQN, Type, PK},
```
ThomasArts, line 600: this code gets computationally much heavier using `to_retail` instead of only `is_full_journalentry`....

src/leveled_iclerk.erl
```diff
@@ -549,44 +584,31 @@ safely_log_filescore(PositionList, FN, Score, SW) ->
...
+                                    FilterFun,
+                                    FilterServer,
+                                    MaxSQN,
+                                    ReloadStrategy),
+                    case ToRetain of
                         true ->
-                            Check = FilterFun(FilterServer, PK, SQN),
-                            case {Check, SQN > MaxSQN} of
-                                {current, _} ->
```
ThomasArts, line 583: the old code can take this path without strategy being "retain" but new code cannot.

ThomasArts (changes requested):
Need to discuss the differences in behaviour for retain once more

ThomasArts (approved):

priv/leveled.schema
martinsumner: Amended - note that the percentage is the inverse of the space released

src/leveled_iclerk.erl
martinsumner: This is true, but the additional check is necessary to resolve the issue, and the extra load is compensated for by the act of caching

PR #322, opened 2020-12-04
is_active check within fold

martinsumner:
Previously every fold had start a clock at the beginning of the fold (to provide a reference time for handle expiring elements), and then pass each key through leveled_codec:is_active/3 before adding to the accumulator.

A more consistent approach would be to set the stop watch when the snapshot is being taken, and apply leveled_codec:is_active as standard within the keyfolder (in maybe_accumulate).

There is an important side effect of this change - improved performance for bucket listing.  If a whole bucket is deleted, in the store there may be lots of tombstones for that bucket and no objects.  With the previous code, every time one of those tombstones was reached the penciller would return to the folder function (as max_keys had been reached), the folder would determine that key is not active, and call again to the penciller to fetch the next key.

The "call again to the penciller to fetch the next key" was an expensive operation, as the penciller's memory has to be re-merged each time.  It is expected to be OK to do that for every bucket, but when there is a bucket full of tombstones, it is very expensive to do that for every key.

Changing the code to place the is_active check within the fold has brought down an example of this query in a load environment (where a bucket of 30M keys has been deleted) from > 60m to under 30s.

### DIFF
--- a/src/leveled_runner.erl
+++ b/src/leveled_runner.erl
@@ -45,8 +45,26 @@
 -type key_range() 
             :: {leveled_codec:ledger_key()|null, 
                     leveled_codec:ledger_key()|null}.
--type fun_and_acc()
-            :: {fun(), any()}.
+-type foldacc() :: any().
+    % Can't currently be specific about what an acc might be
+
+-type fold_objects_fun()
+    :: fun((leveled_codec:key(), leveled_codec:key(), any(), foldacc())
+                -> foldacc()).
+-type fold_keys_fun()
+    :: fun((leveled_codec:key(), leveled_codec:key(), foldacc())
+                -> foldacc()).
+-type fold_buckets_fun()
+    :: fun((leveled_codec:key(), foldacc()) -> foldacc()).
+-type fold_filter_fun()
+    :: fun((leveled_codec:key(), leveled_codec:key()) -> accumulate|pass).
+
+-type snap_fun()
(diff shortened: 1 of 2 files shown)


src/leveled_runner.erl
```diff
@@ -671,49 +644,43 @@ accumulate_objects(FoldObjectsFun, InkerClone, Tag, DeferredFetch) ->
...
-                                        leveled_inker:ink_keycheck(InkerClone,
-                                                                    LK,
-                                                                    SQN),
-                                    case InJournal of
-                                        probably ->
-                                            FoldObjectsFun(B, K, ProxyObj, Acc);
-                                        missing ->
-                                            Acc
-                                    end;
-                                false ->
```
ThomasArts, line 703: This line suggests that JournalCheck always is a boolean value, otherwise this would have crashed... but in line 628 it is suggested that JournalCheck could have a different value than false, otherwise why not doing an "and" and check true and false there?

ThomasArts (approved):
This looks like a nice simplification that cleans up the code.

The difference kicks in when the levelled_codec is not active. I have a bit of trouble to see whether this may happen half-way an ongoing fold and whether the impact is the expected impact.

src/leveled_runner.erl
martinsumner: Clarified this, and added types to help dialyzer going forward

martinsumner:
https://github.com/martinsumner/leveled/pull/325

Issue #321, opened 2020-12-04
List buckets - not always fast


martinsumner:
Bucket listing is intended to be fast.  However, if there are a very large number of tombstones within the penciller, and a large level zero cache - it can still be very expensive.

This is because the query has to be setup each time next_key is called, but as well as being on discovery of a new bucket, this can also be on discover of an inactive key.  So hundreds of thousands of inactive keys requires the query to be re setup (e.g. filtering/merging the levelzero cache, checking the manifest etc) hundreds of thousands of times - if all the keys at the start of the bucket are inactive.  If just one key is active there are no issues.

All folds need to do an is_active check - but they all do the check in the leveled_runner.  The is_active check would be better to be a fundamental part of the keyfolder within the penciller.  so the when trying fetch only a max key of one, it will return the first active key - not every inactive key in turn.

martinsumner:
https://github.com/martinsumner/leveled/pull/320

Issue #319, opened 2020-11-19
Reducing scoring load


martinsumner:
When running many journal compactions per day, and with large journals - there is a lot of overhead scoring and re-scoring files. In some Riak installations journal compaction scoring is dominant factor in disk utilisation.

Couple of potential solutions:

- remember previous individual file scores, and only re-score a random selection of files each run

- when multiple runs are above the target threshold, cache the runs and simply pick one from the cache rather than re-scoring the Journal until the cache is depleted.

The second solution doesn't help when all runs are below the threshold (but if this is a recurring scenario then the number of compactions per day can just be decreased).

The first solution has some issues with functional testing due to its unpredictable nature - i.e. we couldn't be sure that after a bunch of changes the next compaction event will trigger the anticipated compaction.

martinsumner:
As part of this change it is important that the default compaction percentage settings are changed to emphasise the benefit in compacting longer runs.  

For example, if there is a mass deletion event that deletes lots of objects in an adjacent set of journal files - if those files are then scored in separate runs, there might now be an additional tendency to pick each of those files individually, and not as a run.

martinsumner:
https://github.com/martinsumner/leveled/pull/320/commits/00823584eca37c030950d516f16ea1f14958a8a3 to address issue raised above

martinsumner:
https://github.com/martinsumner/leveled/pull/320

PR #323, opened 2020-12-04
Mas i321 checkactive

martinsumner:
Prepare for merge into develop-2.9


martinsumner:
Resolves https://github.com/martinsumner/leveled/issues/321 and https://github.com/martinsumner/leveled/issues/319

CDB files can now cache their last compaction scores, and the inker clerk can randomly choose whether or not to use the caches score.  Issues with scoring and compacting logic diverging also resolves, to prevent a file being scored for compaction where no compaction will actually take place (then getting re-scored etc).

The SST files now build their cache dynamically after startup on folds, not just fetches.  Also they cache the highest LMD for the whole file when the cache is complete.  This accelerates LMD range folds by an order of magnitude.

the active check is now embedded in the fold, rather than being repeatedly embedded in each runner function.  Can significantly speed up bucket listing when there exists buckets with a large number of tombstones at the front of the key range.

PR #342, opened 2021-07-26
Make deletion confirmation an async message

martinsumner:
Currently a sync call is made from cdb to inker when confirming deletion (checking no snapshots have an outstanding requirement to access the file), whereas an async call is made from sst to penciller to achieve the same thing.

Due to the potential of timeouts and crashes - the cdb/inker delete confirmation process is now based on async message passing, making it consistent with the sst/penciller delete confirmation process.


src/leveled_cdb.erl
```diff
@@ -762,26 +769,24 @@ delete_pending({key_check, Key}, _From, State) ->
...
-                                        State#state.filename, 
-                                        State#state.waste_path),
-                    {stop, normal, State};
-                false ->
-                    {next_state,
-                        delete_pending,
-                        State,
-                        ?DELETE_TIMEOUT}
-            end;
+            ok =
```
ThomasArts, line 772: If the process is alive, you may send the async message for each new timeout. That is, if DELETE_TIMEOUT is 1 second and the inker is very busy, then you may send 10 messages in 10 seconds.
This would be impossible before, since you wait for the inker and if the inker is busy, you would timeout the actual call.

ThomasArts (commented):
Make sure the slightly changed semantics of re-sending delete confirmations is what you really want.

src/leveled_cdb.erl
martinsumner: Yes, and any one of those requests leading to a confirmation will cause the leveled_cdb file to close.

PR #344, opened 2021-07-26
Make deletion confirmation an async message - develop31

martinsumner:
Currently a sync call is made from cdb to inker when confirming deletion (checking no snapshots have an outstanding requirement to access the file), whereas an async call is made from sst to penciller to achieve the same thing.

Due to the potential of timeouts and crashes - the cdb/inker delete confirmation process is now based on async message passing, making it consistent with the sst/penciller delete confirmation process.


martinsumner:
Commit from https://github.com/martinsumner/leveled/pull/342

PR #351, opened 2021-08-23
Mas i350 selectivesync

martinsumner:
Add support to determine if a PUT should be sync'd to disk on a case by case basis

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -51,7 +51,7 @@
         book_plainstart/1,
         book_put/5,
         book_put/6,
-        book_put/7,
+        book_put/8,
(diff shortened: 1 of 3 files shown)


martinsumner:
Double-checked about sync during journal compaction.  When we're compacting the journal, is the sync strategy employed (which would be a and and unnecessary thing) should sync be enabled.

Although the iclerk inherits the standard cdb_options, which includes sync_strategy - all writes during journal compaction use cdb_mput not cdb_put, and cdb_mput doesn't check sync_strategy.  So having a strategy of sync does not cause an issue in Journal compaction

ThomasArts (approved):
Adding sync to appropriate places looks good

PR #354, opened 2021-09-30
Search the loader's mock cache ..

martinsumner:
... in the correct direction - otherwise frequently updated objects may not be indexed correctly on reload.


ThomasArts (approved):
Good. This reverses the way in which we find the key in the queue.
Used to look from the back, now from the front. That's the patch needed.

PR #356, opened 2021-10-15
Handle a Filename not being present in pending_deletes

martinsumner:
This may be due to a situation whereby a second confirm_delete has been sent before the sst_deleteconfirmed has been received from the first request.

Now this will now cause inert action rather than crashing.


ThomasArts (approved):
Not sure you want to release the snapshot in case you cannot find the file in the dict, but otherwise perfectly fine refactoring

PR #357, opened 2021-10-18
Handle a Filename not being present in pending_deletes

martinsumner:
This may be due to a situation whereby a second confirm_delete has been sent before the sst_deleteconfirmed has been received from the first request.

Now this will now cause inert action rather than crashing.


martinsumner:
https://github.com/martinsumner/leveled/pull/356

PR #362, opened 2021-10-26
Resolve hash 0 of confusion

martinsumner:
Ensure that a hash of 0 is not confused with an empty index entry (where both hash and position are 0).  For a genuine entry position must always be non-zero.


ThomasArts (commented):
looks good to me.
All tests passed and QuickCheck property passed as well

PR #371, opened 2022-03-11
Mas i370 d30 sstmemory

martinsumner:
On very large stores with lots of old, and largely untouched objects, too much memory is taken up by relatively inactive files. 

The fetch_cache has an unpredictable size, and it is hard to reason or prove its effectiveness.

This change takes the existing caching level configuration, which prevents automatic fadvise of files below a certain level in the LSM tree, and also blocks the creation of the fetch_cache below a certain level.

There has been some observations of memory fragmentation as well.  There is currently no hibernation of unused files.  Unused files will still be used during all object folds and journal compaction because of SQN checking, so SQN checking is now used as a potential trigger for hibernation - if there is no activity in the 60s following. a SQN check the managing process for the SST file will hibernate.

### DIFF
--- a/src/leveled_penciller.erl
+++ b/src/leveled_penciller.erl
@@ -1450,7 +1450,8 @@ roll_memory(State, true) ->
 %% the result tuple includes the level at which the result was found.
 timed_fetch_mem(Key, Hash, Manifest, L0Cache, L0Index, Timings) ->
     SW = os:timestamp(),
-    {R, Level} = fetch_mem(Key, Hash, Manifest, L0Cache, L0Index),
+    {R, Level} =
+        fetch_mem(Key, Hash, Manifest, L0Cache, L0Index, fun timed_sst_get/4),
     UpdTimings = update_timings(SW, Timings, R, Level),
     {R, UpdTimings}.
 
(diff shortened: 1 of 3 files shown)


src/leveled_penciller.erl
```diff
@@ -1528,29 +1538,26 @@ log_slowfetch(T0, R, PID, Level, FetchTolerance) ->
...
+compare_to_sqn(ObjSQN, SQN) when is_integer(ObjSQN), ObjSQN > SQN ->
+    replaced;
+compare_to_sqn(ObjSQN, _SQN) when is_integer(ObjSQN) ->
+    % Normally we would expect the SQN to be equal here, but
+    % this also allows for the Journal to have a more advanced
+    % value. We return true here as we wouldn't want to
+    % compact thta more advanced value, but this may cause
+    % confusion in snapshots.
+    current;
 compare_to_sqn(Obj, SQN) ->
```
ThomasArts, line 1559: Nice refactoring... gets more readable

src/leveled_sst.erl
```diff
@@ -1082,14 +1123,72 @@ member_check(Hash, {sets, HashSet}) ->
...
+        CH -> ExtraHash band (CH - 1)
+    end.
+
+%% @doc
+%% The lower the level, the bigger the memory cost of supporting the cache,
+%% as each level has more files than the previous level.  Load tests with
+%% any sort of pareto distribution show far better cost/benefit ratios for
+%% cache at higher levels.
+-spec cache_size(non_neg_integer()) -> no_cache|4|32|64.
+cache_size(N) when N < 3 ->
```
ThomasArts, line 1162: This works, because `no_cache < 3` but rather:

```suggestion
cache_size(N) when is_integer(N), N < 3 ->
```

src/leveled_sst.erl
```diff
@@ -578,15 +595,19 @@ starting({sst_new,
...
                     Penciller, MaxSQN,
                     OptsSST, IdxModDate}, _From, State) -> 
     DeferredStartupTuple = 
         {RootPath, Filename, Penciller, MaxSQN, OptsSST,
             IdxModDate},
     {reply, ok, starting,
-        State#state{deferred_startup_tuple = DeferredStartupTuple, level = 0}};
+        State#state{
+            deferred_startup_tuple = DeferredStartupTuple,
+            level = 0,
```
ThomasArts, line 609: This will fail... needs
```suggestion
            Level = 0,
```

ThomasArts:
```suggestion
            Level = 0,
```
and then use Level in new_cache?

ThomasArts (approved):

PR #376, opened 2022-05-06
Mas i1826 pollpush

martinsumner:
If there is a backlog in the penciller, but all writes stop, there will be no trigger to push the build-up in-memory to disk.

Now, on journal compaction there will be an additional push of the ledger cache (if it is over the maximum size) that will prompt this as a result of that compaction


martinsumner:
Same concept now included in https://github.com/martinsumner/leveled/pull/382

PR #377, opened 2022-05-18
Mas i370 deletepending

martinsumner:
Previously delete_confirmation was blocked on work_ongoing.

However, if the penciller has a work backlog, work_ongoing may be a recurring problem ... and some files, may remain undeleted long after their use - lifetimes for L0 fails in particular have seen to rise from 10-15s to 5m +.

Letting L0 files linger can have a significant impact on memory.  In put-heavy tests (e.g. when testing riak-admin transfers) the memory footprint of a riak node has bene observed peaking more than 80% above normal levels, when compared to using this patch.

This PR allows for deletes to be confirmed even when there is work ongoing, by postponing the updating of the manifest until the manifest is next returned from the clerk.

### DIFF
--- a/src/leveled_pmanifest.erl
+++ b/src/leveled_pmanifest.erl
@@ -550,7 +550,7 @@ release_snapshot(Manifest, Pid) ->
 %% @doc
 %% A SST file which is in the delete_pending state can check to see if it is
 %% ready to delete against the manifest.
-%% This does not up date the manifest, a call is required to clear_pending to
+%% This does not update the manifest, a call is required to clear_pending to
 %% remove the file from the manifest's list of pending_deletes.
 -spec ready_to_delete(manifest(), string()) -> boolean().
 ready_to_delete(Manifest, Filename) ->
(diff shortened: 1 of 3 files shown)


martinsumner:
The following chart shows memory usage on the joining node over a 15-hour node join process (with transfer-limit of 6) - comparing runs with 3.0.9, 3.0.10 (without this PR) and 3.0.10 (with this PR e.g. +leveled370):

RiakMemory_NodeJoin

To demonstrate the difference, the following chart shows the same three tests, but this time comparing the average lifetime of the L0 SST files (in seconds):
 
L0LifeTime_NodeJoin

src/leveled_pmanifest.erl
```diff
@@ -545,32 +546,41 @@ release_snapshot(Manifest, Pid) ->
                                 min_snapshot_sqn = MinSnapSQN}
     end.
 
--spec ready_to_delete(manifest(), string()) -> {boolean(), manifest()}.
+
 %% @doc
 %% A SST file which is in the delete_pending state can check to see if it is
 %% ready to delete against the manifest.
+%% This does not up date the manifest, a call is required to clear_pending to
```
ThomasArts:
```suggestion
%% This does not update the manifest, a call is required to clear_pending to
```

ThomasArts (approved):
Looks good

Issue #379, opened 2022-10-03
Snapshot time


martinsumner:
In this context, a snapshot of the leveled store is a process which can run a read only query returning results which are fixed at the point in time the snapshot was taken. 

When taking a snapshot of the penciller to run a query, a copy of the penciller manifest is taken (as with all snapshots).  Then as part of the query, in the snapshot penciller process the manifest is slimmed down, to just the part relevant for the query.

There is a Riak user that wishes to run > 10K 2i queries per second.  One of the limiting factors for running high volumes of queries is the snapshot time, as the vnode is blocked  as the  snapshot is taken and every vnode must be involved in 1/N queries.

Currently a snapshot takes approximately 1.2ms.  This limits throughput to (1000/1.2)*N -> 2,500 queries per second with a N-val of 3.

The majority of the time is caused by the copy of manifest (not directly when the copy_manifest function is called, but at the point the copied manifest is returned to the calling process).  The query to provide the slimmed down copy of the manifest is an order of magnitude faster than this copy.

By running the query against the manifest, and then just copying the slimmed down result to the new snapshot penciller, the time taken to perform a snapshot drops from 1200 microseconds, to 200 microseconds.  This changes the throughput limit for 2i queries to This limits throughput to (1000/0.2)*N -> 15,000 queries per second with a N-val of 3.

PR #380, opened 2022-10-03
Query don't copy

martinsumner:
Queries the manifest to avoid copying the whole manifest when taking a snapshot of a penciller to run a query.

Change the logging of fold setup in the Bookie to record the actual snapshot time (rather than the uninteresting and fast returning the the function which will request the snapshot).

A little tidy to avoid duplicating the ?MAX_LEVELS macro.

https://github.com/martinsumner/leveled/issues/379

### DIFF
--- a/src/leveled_penciller.erl
+++ b/src/leveled_penciller.erl
@@ -225,9 +225,13 @@
 
 -record(state, {manifest ::
                     leveled_pmanifest:manifest() | undefined | redacted,
-                query_manifest :: list() | undefined,
+                query_manifest :: 
+                    {list(),
+                        leveled_codec:ledger_key(),
+                        leveled_codec:ledger_key()} | undefined,
                     % Slimmed down version of the manifest containing part
-                    % related to  specific query
+                    % related to  specific query, and the StartKey/EndKey
+                    % used to extract this part
 
                 persisted_sqn = 0 :: integer(), % The highest SQN persisted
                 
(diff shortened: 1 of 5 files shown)


martinsumner:
Testing in standard NHS Riak test cluster on a 2i heavy workload.  The result was that the B0017 log showed a total of 1.173ms to snapshot a store prior to the change, and 233ms after the change.

martinsumner:
With the database unstressed (i.e. just ten bench workers), the throughput increase was 26.6% with a 10% drop in CPU used.

src/leveled_penciller.erl
```diff
@@ -809,21 +795,16 @@ handle_call({fetch_keys,
...
     %% Rename any reference to loop state that may be used by the function
     %% to be returned - https://github.com/martinsumner/leveled/issues/326
-    Manifest = State#state.manifest,
+    SSTiter =
+        case State#state.query_manifest of
+            undefined ->
+                leveled_pmanifest:query_manifest(
+                    State#state.manifest, StartKey, EndKey);
+            QueryManifest ->
+                QueryManifest
```
ThomasArts, line 809: I have a bit of problem to see whether this is compatible with the old solution.

If you register a snapshot, then you now put the query_manifest in the state (line 881). A consecutive fetch_keys will use that query.
But the StartKey and EndKey of the registers snapshot and the fetch might be different? In which case the result of the fetch is different??

More precise, if the fetch is on Query {StartKey = 10, EndKey = 20} and the fetch is on StartKey = 1, EndKey = 8, then the original would use the 1..8 range for fetch, whereas the new one would have saved the query 10...20 in the state and use that. What am I missing?

martinsumner: If at the point of requesting the snapshot you know what the Query is, then the StartKey/EndKey are provided here.  So it is assumed that there will only be one subsequent call to find_keys on that snapshot, and it will be for the same StartKey/EndKey.

There are circumstances where a snapshot is reused for multiple queries (e.g. when doing list buckets), but in this case there is no Query provided when registering the snapshot.

I think it might be worth making the query_manifest a tuple of {SubManifest, StartKey, EndKey} and a guard to crash a subsequent find_keys request which tries a StartKey or EndKey outside of the range.  This will then make this requirement explicit.

ThomasArts (approved):
I like the solution to keep Start and End key in the state!

Issue #370, opened 2022-03-10
Memory utilisation for large stores


martinsumner:
The memory overhead for very large stores, can be very large.

The number of SST files can become very large (e.g. when using Level 6).  There are two potential problems here:

- The memory footprint of each file might be too high;
- There is no help given to the GC (i.e. we don't hibernate rarely used files as happens in the Journal).

Proposal is to:

- For files where we're not pre-loading into page cache, don't have. a fetch_cache  as well.  As the fetch_cache contains Key/MD not just Keys it might be quite large.  It can also be unnecessarily filled by fold_objects or journal_compaction via the  pcl_check_sequencenumber function.

martinsumner:
https://github.com/martinsumner/leveled/pull/371

martinsumner:
Another issue is with work backlogs and L0 files.  L0 files after they're written retain a large number of references to large binaries ... and will commonly be the biggest individual process in the store.  This is OK, as L0 files are generally very short-lived.

However, if there is a work backlog, all SST files can only transition from delete_pending to closed if when delete confirmation is requested to the penciller there is no ongoing work (as during ongoing work the manifest is controlled by the clerk not the penciller).

In pure PUT workloads (such as in riak-admin transfers), this can lead to very high mean durations for L0 files (an in effect lots of L0 files being concurrently alive).  The lifetime of a L0 file will go from 10-20s in high balanced loads, to 300-400s. This can mean there are 20-30 L0 files per db (and potentially 100s per Riak node) that are concurrently stuck in delete_pending.

PR #381, opened 2022-11-10
Mas i370 patch c

martinsumner:
https://github.com/martinsumner/leveled/issues/370

Previous attempts to manage memory usage in large stores has focused on the Block Index Cache, which is about 2 bytes per key (for directly fetchable keys i.e. not index entries) in the leveled_sst files.  These caches grow through use, but were never expunged - and perhaps files with mainly dead data would falsely retain these caches without performance benefits.

The balance is difficult though, between clearing these caches and the non-functional value in  retaining them (especially wrt to increasing the speed of AAE-related queries).

Running a 36 hour test with relatively small values, to build up a key store.  

Three things are noted:

- Memory fragmentation across the nodes on the cluster is between 52% and 56%.
- The memory footprint of leveled_sst processes is still large, even when the Block Index Cache has been expunged (following journal compaction).
- The memory footprint of leveled _sst files (as measured by `process_info(P, memory)`) is commonly about 8 times larger than the footprint of process state (as measured by `erts_debug:flat_size(sys:get_state(P))`) even after forced garbage collection on the PID.  It is some times 25 times larger.  This is the case even when binary memory owned by the process is empty (as measured by `process_info(P, binary)`). **EDIT ... IGNORE THIS! `erts_debug:flat_size/1` is measured in words not bytes!  One word equals 8 bytes on a 64-bit system**

This PR is a combination of three refactoring exercises intended to try and improve this situation:

- https://github.com/martinsumner/leveled/commit/d393c99856ba79dff2da0b83157f43eb4e9cb6e2 which refactors the penciller memory.  Partly for readability (to stop LoopState being passed to internal functions), but also to try and do memory efficient operations where possible (e.g. `[Addition|ExistingList]` to extend lists).
- https://github.com/martinsumner/leveled/commit/415e682c6d8d2b9c9354a05f487abd2984e4dc44 which changes the index kept within a leveled_sst file (of 1 in 128 keys).  This now attempts to find common elements in this index, (e.g. where all entries in the SST file have the same Type/Bucket) and then only index the deltas.
- https://github.com/martinsumner/leveled/commit/56c5e2565ddde15edebd7c0097cb57882b91d3ed which fixes a bug in the Block Index Cache which meant that when a BIC is cleared, if memory of the HMD date for the file is retained, the cache would never be rebuilt. The refactoring also reduces unnecessary changes to the BIC as the BIC is populated, and an unnecessary fold over the BIC to rediscover the HMD.

It is hoped that tidying up these elements will make it easier to investigate further memory overheads (and perhaps in the case of the leveled_sst index changes, directly improve memory footprint).

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -162,7 +162,7 @@
                         loader = leveled_tree:empty(?CACHE_TYPE)
                                     :: tuple()|empty_cache,
                         load_queue = [] :: list(),
-                        index = leveled_pmem:new_index(), % array or empty_index
+                        index = leveled_pmem:new_index(),
                         min_sqn = infinity :: integer()|infinity,
                         max_sqn = 0 :: integer()}).
 
@@ -2384,7 +2384,7 @@ addto_ledgercache({H, SQN, KeyChanges}, Cache, loader) ->
 %% Check the ledger cache for a Key, when the ledger cache is in loader mode
 %% and so is populating a queue not an ETS table
 check_in_ledgercache(PK, Hash, Cache, loader) ->
-    case leveled_pmem:check_index(Hash, Cache#ledger_cache.index) of
+    case leveled_pmem:check_index(Hash, [Cache#ledger_cache.index]) of
         [] ->
             false;
         _ ->
--- a/src/leveled_codec.erl
+++ b/src/leveled_codec.erl
@@ -80,6 +80,8 @@
         tomb|{active, non_neg_integer()|infinity}.
 -type ledger_key() :: 
         {tag(), any(), any(), any()}|all.
+-type slimmed_key() ::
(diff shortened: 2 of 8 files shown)


martinsumner:
Resolving my words vs bytes confusion with `erts_debug:flat_size/1`, it does look clear that the majority of the process memory for each leveled_sst file is consumed with the summary index table (rather than the block index cache).  So finding efficiency here is the key, and this commit should have a significant impact - https://github.com/martinsumner/leveled/commit/415e682c6d8d2b9c9354a05f487abd2984e4dc44.

As well as removing redundant/repeated terms, it also reduces the number of index entries that are promoted up to the gb_trees part of the index (and are hence duplicated).

martinsumner:
Replaced by https://github.com/martinsumner/leveled/pull/383

PR #383, opened 2022-11-16
Mas i370 patch d

martinsumner:
https://github.com/martinsumner/leveled/issues/370

Previous attempts to manage memory usage in large stores has focused on the Block Index Cache, which is about 2 bytes per key (for directly fetchable keys i.e. not index entries) in the leveled_sst files.  These caches grow through use, but were never expunged - and perhaps files with mainly dead data would falsely retain these caches without performance benefits.

The balance is difficult though, between clearing these caches and the non-functional value in  retaining them (especially wrt to increasing the speed of AAE-related queries).

Running a 36 hour test with relatively small values, to build up a key store.  

Three things are noted:

- Memory fragmentation across the nodes on the cluster is between 52% and 56%.
- The memory footprint of leveled_sst processes is still large, even when the Block Index Cache has been expunged (following journal compaction).
- The memory footprint of leveled _sst files (as measured by `process_info(P, memory)`) is commonly about 8 times larger than the footprint of process state (as measured by `erts_debug:flat_size(sys:get_state(P))`) even after forced garbage collection on the PID.  It is some times 25 times larger.  This is the case even when binary memory owned by the process is empty (as measured by `process_info(P, binary)`). **EDIT ... IGNORE THIS! `erts_debug:flat_size/1` is measured in words not bytes!  One word equals 8 bytes on a 64-bit system**

This PR is a combination of three refactoring exercises intended to try and improve this situation:

- https://github.com/martinsumner/leveled/commit/d393c99856ba79dff2da0b83157f43eb4e9cb6e2 which refactors the penciller memory.  Partly for readability (to stop LoopState being passed to internal functions), but also to try and do memory efficient operations where possible (e.g. `[Addition|ExistingList]` to extend lists).
- https://github.com/martinsumner/leveled/commit/415e682c6d8d2b9c9354a05f487abd2984e4dc44 which changes the index kept within a leveled_sst file (of 1 in 128 keys).  This now attempts to find common elements in this index, (e.g. where all entries in the SST file have the same Type/Bucket) and then only index the deltas.
- https://github.com/martinsumner/leveled/commit/56c5e2565ddde15edebd7c0097cb57882b91d3ed which fixes a bug in the Block Index Cache which meant that when a BIC is cleared, if memory of the HMD date for the file is retained, the cache would never be rebuilt. The refactoring also reduces unnecessary changes to the BIC as the BIC is populated, and an unnecessary fold over the BIC to rediscover the HMD.

It is hoped that tidying up these elements will make it easier to investigate further memory overheads (and perhaps in the case of the leveled_sst index changes, directly improve memory footprint).


martinsumner:
Extend Patch C to ensure that both an excess bookie ledger cache, and an over-size penciller in-memory cache will be persisted to disk even where no traffic is present

martinsumner:
This compares different configurations on a high-key-volume load test (one using small objects and lots of PUTs to try and rapidly increase the volume of keys).

Memory -

Riak310PatchD_Memory

Throughput -

Riak310PatchD_Throughput

It seems that we get improvements in memory efficiency almost for free in terms of throughput.

The Profile A refers to setting eheap SBCT to 128KB (default 512KB): erlang.eheap_memory.sbct = 128

martinsumner:
The memory is as reported by the OS, averaged across the nodes in the cluster.  The test duration was 36 hours

src/leveled_sst.erl
```diff
@@ -1725,20 +1764,100 @@ hmac(Int) when is_integer(Int) ->
...
+-spec term_prefix_filter(
+    pos_integer(), binary()) ->
+        fun((leveled_codec:ledger_key()) -> leveled_codec:slimmed_key()).
+term_prefix_filter(N, Prefix) ->
+    fun({_Tag, _Bucket, {_Field, Term}, Key}) ->
+        case Term of
+            T when byte_size(T) =< N ->
+                null;
+            <<QueryKey:N/binary, Suffix/binary>> ->
+                case QueryKey of
```
yorkshire-steve: Should this be QueryTerm as opposed to QueryKey for clarity?

src/leveled_sst.erl
```diff
@@ -1725,20 +1764,100 @@ hmac(Int) when is_integer(Int) ->
...
+
+-spec get_filterfun(
+    leveled_codec:ledger_key(), leveled_codec:ledger_key()) ->
+        fun((leveled_codec:ledger_key())
+            -> leveled_codec:ledger_key()|leveled_codec:slimmed_key()).
+get_filterfun(
+        {Tag, Bucket, {Field, FT}, FK}, {Tag, Bucket, {Field, LT}, LK})
+            when is_binary(FT), is_binary(FK), is_binary(LT), is_binary(LK) ->
+    case {binary:longest_common_prefix([FT, LT]), byte_size(FT)} of
+        {N, M} when N > 0, M > N ->
```
yorkshire-steve: I keep looking at this and wondering if there could ever be a situation where FT == LT

martinsumner: If there are a lot of keys with exactly the same term on an index, then yes.  That needs a unit test

yorkshire-steve: In that case when FT==LT, won't M==N? Then the test M > N is false and this won't use the more optimised FilterFun

martinsumner: That is correct.  So it works functionally (as it switches to using {Term, Key} on the index), but is not optimised.  I've added a unit test to show that it at least works.

I don't think changing M > N to M >= N will work, so we would need another clause to optimise.

martinsumner: See https://github.com/martinsumner/leveled/pull/383/commits/9db486671cf5b2576ea33fb5d1dbd639c499d7aa

martinsumner:
Tests run to include the latest version of PatchD, and also an additional control on OTP 25 (without leveled fixes):

Riak310PatchD_Memory2

The positive news is increased memory efficiency in OTP 25.  However, this doesn't deliver the same level of improvements as the patch

src/leveled_sst.erl
martinsumner: I think the interim assignment to anything is actually unnecessary.  We actually only have two cases, where it maps to <<Prefix:N/binary, Suffix/binary>> ... or where the result should be null.  This is the same in the key_prefix_filter/2 and the term_prefix_filter/2.

martinsumner: https://github.com/martinsumner/leveled/pull/383/commits/76ed44dea489d1fd009d9d6123d92199deb7a72a

yorkshire-steve: I think this change allows `Suffix` to be null (when Key == Prefix), which wasn't allowed previously due to `K when byte_size(K) =< N`
I don't think this will cause a problem though as the return Suffix would be null which is what was being returned previously
Just wanted to point it out in case you knew of any unintended side effect here

martinsumner: In this case the Suffix would be <<>> and not null I think.  However, this is a good point.  Let me think

src/leveled_sst.erl
```diff
@@ -1725,20 +1764,107 @@ hmac(Int) when is_integer(Int) ->
...
+    end;
+get_filterfun(
+        {Tag, Bucket, FK, null}, {Tag, Bucket, LK, null})
+            when is_binary(FK), is_binary(LK), FK < LK ->
+    case {binary:longest_common_prefix([FK, LK]), byte_size(FK)} of
+        {N, M} when N > 0, M > N ->
+            <<Prefix:N/binary, _Rest/binary>> = FK,
+            key_prefix_filter(N, Prefix);
+        _ ->
+            fun key_filter/1
```
yorkshire-steve: I can't quite remember the rules on ordering, but is it possible to have FK="0000" and LK="0000999999"? or does "0000" come after? if it's the first, then I don't think this would result in an optimised FilterFun. In the 2nd case, this may need a test case to ensure expected behaviour (there may be one already I've not spotted)

martinsumner: So <<"0000">> is smaller than <<"0000">> with anything appended so <<"0000999999">> is > than <<"0000">>.

I think the fact that not every case is optimised is fine, it isn't going to have a material impact if the occasional SST file is using more memory than is strictly required.  So I would rather have simpler code than have full optimisation.

The unit test `key_matchesprefix_test/0` In both the index case and object case creates a First Key that is equal in size to the longest_common_prefix.  But I perhaps added this test as I assumed it would be optimised, when no in fact it just becomes a `key_filter/1` test.

Need to clear my head of other things, and just think this through carefully again.

martinsumner: I'm sure M == N is ok.  The guard M > N was introduced before the Prefix check was added to the filter functions (we were just truncating for size).  The weirdness that might have been caused in these cases is now handled by setting to null whenever the first N don't match the Prefix.

I've updated in https://github.com/martinsumner/leveled/pull/383/commits/0fe735fa823daf9f92e2d468783fbec64483bd49 and all tests pass.

martinsumner:
Further test.  This starts from a cluster pre-populated with 360M objects, and then aggressively inserts more over 24 hours (with load shifting to be more read-based in the last hour).  

Memory footprint:

Riak310PatchD_Memory3

Throughput achieved:

Riak310PatchD_Throughput3

The PatchD performs well in terms of both throughput and memory optimisation.  Adding the alternative configuration profile (sbct = 128KB ), reduces throughput at the start of the tense - but consistently delivers a lower memory footprint.

A longer test needs to be run from the same base, to understand the longer-term behaviour when load goes primarily GETs not PUTs.

yorkshire-steve (approved):

PR #385, opened 2022-12-05
Mas i370 patch e

martinsumner:
Refactor the  approach to logging within leveled.  There are two primary changes:

- Move gathering of request-related statistics to a dedicated process (leveled_monitor).
- Refactor leveled_log to improve efficiency, switching ?LOGBASE to a map based on atoms/binary, and improving the efficiency of timestamp printing (borrowing from erlang-lager).
- Add some equality checks so that LoopState is now only updated when necessary.


martinsumner:
Testing of memory and throughput with large key counts (and index entries).  Approx 350M rising to 650M objects during the test, with approx 2bn index entries.

Throughput:

Riak310PatchE_Throughput4

Memory:

Riak310PatchE_Memory4

ProfileA sets sbct (eheap) to 128KB.  ProfileB sets sbct (eheap) to 256KB.  Control is Riak 3.0.11,

martinsumner:
Re-tested with the standard NHS load test.  This generates a lower volume of objects, and uses objects of different size.  However, even then the memory difference between 3.0.11 and Riak+Patch E is notable.  Riak + Patch E +sbct to 128KB uses less memory still.  Differences in throughput between configurations was < 1%

Riak310PatchE_Memory5

ThomasArts (approved):
Code review of leveled_monitor
Looks good.

martinsumner:
Reviewed with @azharnisar - checked new logs looked AOK

PR #395, opened 2023-02-20
Initial support for OTP 26

martinsumner:
Add support for OTP 26.

Fixes some type issues newly discovered by improved dialyzer, and also keeps dialyzer happy by moving functions and includes only used in test, into the ifdef(TEST).

### DIFF
--- a/src/leveled_cdb.erl
+++ b/src/leveled_cdb.erl
@@ -112,8 +112,6 @@
 -export([finished_rolling/1,
             hashtable_calc/2]).
 
--include_lib("eunit/include/eunit.hrl").
-
 -define(DWORD_SIZE, 8).
 -define(WORD_SIZE, 4).
 -define(MAX_FILE_SIZE, 3221225472).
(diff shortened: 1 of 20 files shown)


src/leveled_cdb.erl
```diff
@@ -1883,10 +1895,12 @@ create(FileName,KeyValueList) ->
 
 %% Should not be used for non-test PUTs by the inker - as the Max File Size
 %% should be taken from the startup options not the default
-put(FileName, Key, Value, {LastPosition, HashTree}) ->
-    put(FileName, Key, Value, {LastPosition, HashTree},
-            ?BINARY_MODE, ?MAX_FILE_SIZE, false).
-
+put(FileName, Key, Value, {LastPosition, HashTree}) when is_list(FileName) ->
+    file_put(FileName, Key, Value, {LastPosition, HashTree},
```
ThomasArts, line 1899: using `file_put`here and therefore correct

src/leveled_iclerk.erl
```diff
@@ -863,7 +861,7 @@ filter_output_fun(FilterFun, FilterServer, MaxSQN, Strategy) ->
 
 -spec to_retain(leveled_codec:journal_key(),
                 leveled_inker:filterfun(),
-                leveled_inker:fillter_server(),
+                leveled_inker:filterserver(),
```
ThomasArts, line 864: nice catch... is that improved dialyzer?

ThomasArts (approved):
Looks good to me

src/leveled_iclerk.erl
martinsumner: Yes. OTP26 now seems to detect lots of "obvious" problems that weren't revealed before - in particular mis-typed type references, and references to unexported types.

PR #399, opened 2023-02-21
Spawn to deserialise blocks offline

martinsumner:
Hypothesis is that the growth in the heap necessary due to continual term_to_binary calls to deserialise blocks is wasting memory - so do this memory-intensive task in a short-lived process.

### DIFF
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -434,7 +434,8 @@ sst_get(Pid, LedgerKey) ->
 %% Return a Key, Value pair matching a Key or not_present if the Key is not in
 %% the store (with the magic hash precalculated).
 sst_get(Pid, LedgerKey, Hash) ->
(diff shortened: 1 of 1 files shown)


src/leveled_sst.erl
```diff
@@ -2164,8 +2164,11 @@ check_blocks([Pos|Rest], BlockPointer, BlockLengths, PosBinLength,
                     PosBinLength,
                     BlockNumber,
                     additional_offset(IdxModDate)),
-    R = fetchfrom_rawblock(BlockPos, deserialise_block(BlockBin, PressMethod)),
-    case {R, LedgerKeyToCheck} of
+    Pid =
+        spawn(
+            ?MODULE, check_block, [self(), BlockPos, BlockBin, PressMethod]),
+    Result = receive {checked_block, Pid, R} -> R end,
```
ThomasArts: This is just code to verify an idea... otherwise, we should have used some more caution here on timeout of a possible crash of spawned code, etc.... but not for this experiment.

PR #401, opened 2023-02-28
Refactored leveled_sst gen_fsm into gen_statem

ThomasArts:


src/leveled_cdb.erl
```diff
@@ -386,63 +367,60 @@ cdb_deletepending(Pid, ManSQN, Inker) ->
...
 
 -spec cdb_clerkcomplete(pid()) -> ok.
 %% @doc
 %% When an Inker's clerk has finished with a CDB process, then it will call
 %% complete.  Currently this will prompt hibernation, as the CDB process may
-%% not be needed for a period.  
+%% not be needed for a period.
 cdb_clerkcomplete(Pid) ->
-    gen_fsm:send_all_state_event(Pid, clerk_complete).
+    gen_statem:cast(Pid, clerk_complete).
```
martinsumner, line 411: The clerk_complete message can only be sent when in the `reader` state.  It is send only by the leveled_iclerk, and the leveled_iclerk only examines files in this state.  The state cannot be transitioned to delete_pending without the leveled_iclerk first having called clerk_complete - so there should be no race.  So handling this message should be done in the `reader` state only.

There is an issue otherwise with code coverage, in that there will no longer be 100% code coverage (and so spotting regression of code coverage will be hard).

src/leveled_cdb.erl
```diff
@@ -489,48 +470,56 @@ starting({open_writer, Filename}, _From, State) ->
...
-    {reply,
+    {next_state, reader, State0,
+     [{reply, From, ok}, hibernate]};
+starting({call, From}, Event, State) ->
+    handle_sync_event(Event, From, State);
+starting(cast, clerk_complete, _State) ->
+    {keep_state_and_data, [hibernate]};
+starting(info, _Msg, _State) ->
+    {keep_state_and_data, []}.
+
```
martinsumner, line 501: If there are no `info` messages supported should we remove handling of them.  Again code coverage being the issue.

src/leveled_sst.erl
```diff
@@ -680,39 +669,43 @@ starting(complete_l0startup, State) ->
...
                 end,
             FetchFun(length(FetchedSlots) + 1, ReturnFun),
-            {next_state,
-                starting,
+            {keep_state,
                 State#state{new_slots = FetchedSlots}}
-    end.
+    end;
+starting(cast, {update_blockindex_cache, BIC}, State) ->
+    handle_update_blockindex_cache(BIC, State);
```
martinsumner, line 701: This message cannot be received in the startup state, as it is prompted from get_slots which can only occur in the reader state

src/leveled_sst.erl
```diff
@@ -895,41 +885,75 @@ delete_pending({get_kv, LedgerKey, Hash}, _From, State) ->
...
+delete_pending(cast, close, State) ->
     leveled_log:log(sst07, [State#state.filename]),
     ok = file:close(State#state.handle),
     ok = file:delete(filename:join(State#state.root_path,
                                     State#state.filename)),
-    {stop, normal, ok, State}.
+    {stop, normal, State};
+
+delete_pending(cast, {update_blockindex_cache, BIC}, State) ->
+    handle_update_blockindex_cache(BIC, State);
```
martinsumner, line 928: This is interesting. get_slots may be prompted which results in update_blockindex_cache but a set_for_delete message may happen in-between.  Hence we do need support for update_blockindex_cache in delete_pending state - but this is never triggered in test, so we perhaps need a  unit test to explicitly trigger it.

martinsumner (commented):
There are some code coverage issues.  I'm loathe to give up on 100% code coverage, as I'm worried about missing a regression in code coverage if we tolerate anything less.  There is some code here which I think can never be triggered, but also one case (update_blockindex_cache being cast in a race in a switch to delete_pending) which probably needs a test to get coverage.

The casting of timeout sin leveled_sst I think will end when we merge in the latest fix from develop-3.0

martinsumner:
Replaced by https://github.com/martinsumner/leveled/pull/404

PR #403, opened 2023-03-06
Mas i1820 offlinedeserialisation cbo

martinsumner:
Use of pro-active hibernate, hibernate_after and offline processing to reduce fluctuation in memory footprints of leveled_sst processes

### DIFF
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -434,7 +434,8 @@ sst_get(Pid, LedgerKey) ->
 %% Return a Key, Value pair matching a Key or not_present if the Key is not in
 %% the store (with the magic hash precalculated).
 sst_get(Pid, LedgerKey, Hash) ->
-    gen_fsm:sync_send_event(Pid, {get_kv, LedgerKey, Hash}, infinity).
+    gen_fsm:sync_send_event(
+        Pid, {get_kv, LedgerKey, Hash, undefined}, infinity).
 
 -spec sst_getsqn(pid(),
     leveled_codec:ledger_key(),
@@ -443,7 +444,8 @@ sst_get(Pid, LedgerKey, Hash) ->
 %% Return a SQN for the key or not_present if the key is not in
 %% the store (with the magic hash precalculated).
 sst_getsqn(Pid, LedgerKey, Hash) ->
-    gen_fsm:sync_send_event(Pid, {get_sqn, LedgerKey, Hash}, infinity).
+    gen_fsm:sync_send_event(
+        Pid, {get_kv, LedgerKey, Hash, fun sqn_only/1}, infinity).
 
 -spec sst_getmaxsequencenumber(pid()) -> integer().
 %% @doc
@@ -708,25 +710,10 @@ starting({sst_returnslot, FetchedSlot, FetchFun, SlotCount}, State) ->
                 State#state{new_slots = FetchedSlots}}
     end.
 
-reader({get_sqn, LedgerKey, Hash}, _From, State) ->
+
+reader({get_kv, LedgerKey, Hash, Filter}, _From, State) ->
     % Get a KV value and potentially take sample timings
-    {Result, _BIC, _HMD, _FC} = 
-        fetch(
-            LedgerKey, Hash,
-            State#state.summary,
-            State#state.compression_method,
-            State#state.high_modified_date,
-            State#state.index_moddate,
-            State#state.filter_fun,
-            State#state.blockindex_cache,
-            State#state.fetch_cache,
-            State#state.handle,
-            State#state.level,
-            {no_monitor, 0}),
-    {reply, sqn_only(Result), reader, State, ?HIBERNATE_TIMEOUT};
-reader({get_kv, LedgerKey, Hash}, _From, State) ->
-    % Get a KV value and potentially take sample timings
-    {Result, BIC, HMD, FC} = 
+    {KeyValue, BIC, HMD, FC} = 
         fetch(
             LedgerKey, Hash,
             State#state.summary,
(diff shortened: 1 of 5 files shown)


martinsumner:
RiakMemory_HighKeyCountTest

Comparison of memory used in Riak test designed to drive very high key counts relatively quickly.  The Control is Riak 3.0.13 (with default beam memory settings e.g. eheap memory sbct of 512KB).

martinsumner:
Riak315xlarge_memcomparison

Throughput comparison on 8-node im4gn.xlarge 24 hour test.  Throughput difference was a 2% drop with this patch.  However, there was a 10% memory drop with both running eheap memory sbct of 128KB.

martinsumner:
Important that there is a positive change with and without default SBCT setting.  Changing the default SBCT setting improves memory usage, but in some cases this improvement can be hard to detect because of the use of MADVISE_FREE not MADVISE_DONTNEED.  

Ideally it would be helpful to have a fully observable drop in memory.  Whereas with sbct=128KB when the single block carriers are shrunk with MADVISE_FREE this is not easily observable.

martinsumner:
@ThomasArts @stephenwood24-nhs

src/leveled_sst.erl
```diff
@@ -708,25 +712,10 @@ starting({sst_returnslot, FetchedSlot, FetchFun, SlotCount}, State) ->
...
-            State#state.summary,
-            State#state.compression_method,
-            State#state.high_modified_date,
-            State#state.index_moddate,
-            State#state.filter_fun,
-            State#state.blockindex_cache,
-            State#state.fetch_cache,
-            State#state.handle,
-            State#state.level,
-            {no_monitor, 0}),
```
ThomasArts, line 725: This looks like a pretty big step. On the one hand, you now introduce different monitor options for get_sqn (it used to be `{monitor, 0}` when you fethc, you change it to monitor that is in the state. Probably better, but a change.

On the other hand, you now handle many more alternatives in updating the state after this event. Most remarkable the blockindex_cache update. See commengt below

src/leveled_sst.erl
```diff
@@ -739,17 +728,27 @@ reader({get_kv, LedgerKey, Hash}, _From, State) ->
...
+                KeyValue;
+            F ->
+                F(KeyValue)
+        end,
     case {BIC, HMD, FC} of
         {no_update, no_update, no_update} ->
             {reply, Result, reader, State};
         {no_update, no_update, FC} ->
             {reply, Result, reader, State#state{fetch_cache = FC}};
+        {BIC, undefined, no_update} ->
```
ThomasArts, line 750: This case is new, most likely added for get_sqn case, but it adds a new possibility for get_kv. It probably never was the case that HDM was undefined here... but for a reviewer it is very hard to see the impact of those cases that are new.

src/leveled_sst.erl
```diff
@@ -2664,8 +2657,8 @@ fetch_value([Pos|Rest], BlockLengths, Blocks, Key, PressMethod) ->
     {BlockNumber, BlockPos} = revert_position(Pos),
     {Offset, Length} = block_offsetandlength(BlockLengths, BlockNumber),
     <<_Pre:Offset/binary, Block:Length/binary, _Rest/binary>> = Blocks,
-    RawBlock = deserialise_block(Block, PressMethod),
-    case fetchfrom_rawblock(BlockPos, RawBlock) of 
+    R = fetchfrom_rawblock(BlockPos, deserialise_block(Block, PressMethod)),
```
ThomasArts, line 2668: this could be a spawn_check_block as well.... but chosen not to?

src/leveled_sst.erl
```diff
@@ -843,16 +842,6 @@ reader(close, _From, State) ->
     ok = file:close(State#state.handle),
     {stop, normal, ok, State}.
 
-reader(timeout, State) ->
```
ThomasArts, line 846: I am very surprised you can delete this timeout handling!

There are still a DELETE_TIMEOUT and STARTUP_TIMEOUT that would kick in now and then and that would cause this code to execute. So why does this work when no timeout handling in reader any more?

ThomasArts (commented):
Timeout discussion necessary.

src/leveled_sst.erl
```diff
@@ -2180,6 +2160,19 @@ check_blocks([Pos|Rest], BlockPointer, BlockLengths, PosBinLength,
                             Acc)
     end.
 
+-spec spawn_check_block(non_neg_integer(), binary(), press_method())
+        -> not_present|leveled_codec:ledger_kv().
+spawn_check_block(BlockPos, BlockBin, PressMethod) ->
+    Pid =
```
ThomasArts, line 2174: Parent = self(),
Pid = spawn_link(fun() -> c heck_block(Parent, BlockPos, BlockBin, Pressmethod) end),
receive ...

src/leveled_sst.erl
martinsumner: DELETE_TIMEOUT only happens in the delete_pending state.  The STARTUP_TIMEOUT will prompt a message `start_complete` not a timeout

src/leveled_sst.erl
martinsumner: Thee was a volume test comparing the options of spawning in both cases or just the other case - and there was no benefit from spawning here.  In this case the Block Index Cache is not yet built, and once that cache is built there will be a hibernate - so we wait for the hibernate instead.

src/leveled_sst.erl
martinsumner: So if the BlockIndexCache is built prior to the request, there will be no_update to either BIC or HMD.  If not, and the BIC is still not complete, there will be an update only to the BIC (and HMD will not yet be defined, as it needs to have all members of the BIC to generate the HMD) - and so HMD is undefined.

Only in the case where BIC was not complete before the request, but the request has prompted the BIC to be completed, will there also be a HMD (which can now be calculated).  As the BIC will no longer be updated,  this is the moment chosen to hibernate.

ThomasArts (approved):

src/leveled_sst.erl
martinsumner: This has been changed so that get_sqn will still not be subject to monitor

PR #404, opened 2023-03-13
Mas p401 coverage

martinsumner:
Combines:

https://github.com/martinsumner/leveled/pull/401
https://github.com/martinsumner/leveled/pull/403

This now has 100% test coverage as well.

### DIFF
--- a/test/end_to_end/recovery_SUITE.erl
+++ b/test/end_to_end/recovery_SUITE.erl
@@ -68,6 +68,7 @@ replace_everything(_Config) ->
         testutil:put_altered_indexed_objects(Book1, BKT, KSpcL1),
     ok = testutil:check_indexed_objects(Book1, BKT, KSpcL2, V2),
     compact_and_wait(Book1, 1000),
+    compact_and_wait(Book1, 1000),
     {ok, FileList1} =  file:list_dir(CompPath),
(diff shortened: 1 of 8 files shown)


test/end_to_end/recovery_SUITE.erl
```diff
@@ -68,6 +68,7 @@ replace_everything(_Config) ->
         testutil:put_altered_indexed_objects(Book1, BKT, KSpcL1),
     ok = testutil:check_indexed_objects(Book1, BKT, KSpcL2, V2),
     compact_and_wait(Book1, 1000),
+    compact_and_wait(Book1, 1000),
```
ThomasArts, line 71: Surprising, need to wait one more second now?

ThomasArts (commented):
Looks good to me.
Only remark is the fact that if a cast or info message comes in at specific states, then this is now resulting in a case clause exception, whereas it used to be handled some way.

test/end_to_end/recovery_SUITE.erl
martinsumner: The test was that after one loop there is no further compaction work to be done.  However, there is some potential randomness here, and so sometimes there may have still been viable compaction work for a second loop.  so now we test all compaction work is done in two attempts, not just one.

PR #408, opened 2023-07-12
Use backwards compatible term_to_binary

martinsumner:
So that where we have hashed term_to_binary output in OTP25 or earlier, that has will be matched in OTP 26.

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -702,6 +702,7 @@ book_returnfolder(Pid, RunnerType) ->
                                 when Bucket::term(),
                                      StartKey::term(),
                                      FoldFun::fun((Bucket, Key | {IndexVal, Key}, Acc) -> Acc),
+                                     Key::term(),
                                      Acc::term(),
                                      IndexField::term(),
                                      IndexVal::term(),
--- a/src/leveled_codec.erl
+++ b/src/leveled_codec.erl
@@ -202,7 +202,7 @@ headkey_to_canonicalbinary({?HEAD_TAG, {BucketType, Bucket}, Key, SubKey})
 headkey_to_canonicalbinary(Key) when element(1, Key) == ?HEAD_TAG ->
     % In unit tests head specs can have non-binary keys, so handle
     % this through hashing the whole key
-    term_to_binary(Key).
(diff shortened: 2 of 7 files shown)


ThomasArts (commented):
It is hard to see if you covered all cases.

I did a grep on term_to_binary to see if there might be places that you could have forgotten... One suspect:

[src/leveled_inker.erl:    ManifestBin = term_to_binary(Manifest),
A manifest has a key as last element of the tuple:](https://github.com/martinsumner/leveled/blob/bc87273c76cd7fb86c40a528cbcc1a8cd8d9eb61/src/leveled_inker.erl#L1331C1-L1335C44)
According to the code in leveled_cdb:
https://github.com/martinsumner/leveled/blob/bc87273c76cd7fb86c40a528cbcc1a8cd8d9eb61/src/leveled_cdb.erl#L824C1-L825C66
the last key may have value `empty`. 

So this works fine when you write to disk in OTP25 and then open in OTP26, reading with binary_to_term... however, if OTP26 writes it, you won't be able to open it again with the OTP25 version.

So, there might be some smaller issues left after this refactoring.

martinsumner:
I'm not sure that is a problem.  As I understand it the binary_to_term/term_to_binary is still backwards/forwards compatible between OTP 25 and OTP 26 (as OTP 25 already supported the unicode encoding of atoms, it just wasn't the default behaviour).  The only concern is where we:

- Use term_to_binary;
- Hash the result of using term_to_binary;
- Persist the result hash as part of file structure (e.g. in a bloom filter) so that we expect the hash to match any future request.

In this case if we save the file in OTP 25, and then open when we're running OTP 26, it will trust the bloom filter that it reads off disk.  If OTP 26 then hashes the same key (after converting to binary) and we check for presence of the key in the bloom filter, if OTP 26 hash has a subtle change in binary_to_term, then the hash will be different, and the code will assume the key was not present (due to bloom filter mismatch) when actually it is.

So as long as there is always a hash involved we should be OK, as long as too big a jump in OTP version.  I think the new encoding in OTP 26 can be read using term_to_binary as far back as OTP R16B.

ThomasArts (approved):

Issue #410, opened 2023-09-25
Snapshot releasing on shutdown


martinsumner:
When a bookie is shutdown, there may be ongoing snapshots (supporting queries) which will be impacted by the shutdown.

The penciller and the inker take different approaches.  The penciller shuts down regardless of snapshots, and the inker will close snapshots before shutting down.

Both situations have led to issues. 

- With the inker a snapshot may have been released and shutdown during the shutdown phase of the inker.  This is particularly true when Journal compaction is running, as the clerk must stop before the snapshots are shutdown, and the clerk may be busy scoring/re-writing a file.  This causes the inker to have an unhandled noproc exception when it tries to shutdown the snapshot - and the inker then crashes rather than shutting down.
- When leveled is being run as a parallel key store, aae_folds may still be running when a key store is shutdown due to rebuilds.  This leads to aae_fold query crashes.

martinsumner:
One possible solution is to have an option to pass a delay budget into the calls to book_destroy/book_close so that on shutdown the inker or penciller can loop waiting for snapshots to be released up to the delay budget before triggering shutdown.  The delay budget should be generally small (e.g. 20 seconds) but could be set to a larger value (e.g. 300 seconds) when shutting down a store due to aae_rebuild.

martinsumner:
The level of additional complexity here needs to be balanced against the fairly limited benefit.  The issue with Inker shutdown is an issue in Riak when trying to shut down a node when significant AAE activity is ongoing.  The consequences of this are handled though, eventually due to AAE tree rebuilds triggered on startup.

Likewise the occasional failure of an aae_fold is an inconvenience rather than a critical problem.

PR #411, opened 2023-09-27
Close in stages - waiting for releases

martinsumner:
[Related issue](https://github.com/martinsumner/leveled/issues/410)

Have a consistent approach to closing the inker and the penciller - so that the close can be interrupted by releasing of snapshots.  Then any unreleased snapshots are closed before shutdown - with a 10s pause to give queries a short opportunity to finish.

This should address some issues, primarily seen (but very rarely) in test whereby post-rebuild destruction of parallel AAE keystores cause the crashing of aae_folds.

The primary benefit is to stop an attempt to release a snapshot that has in fact already finished, from causing a crash of the database on normal stop - which in Riak may prevent other closing actions from completing normally.  This was primarily an issue when shutdown is delayed by an ongoing journal compaction job.

### DIFF
--- a/src/leveled_inker.erl
+++ b/src/leveled_inker.erl
@@ -287,6 +287,18 @@ ink_confirmdelete(Pid, ManSQN, CDBpid) ->
 ink_close(Pid) ->
     gen_server:call(Pid, close, infinity).
 
+-spec ink_snapclose(pid()) -> ok.
+%% @doc
+%% Specifically to be used when closing snpashots on shutdown, will handle a
+%% scenario where a snapshot has already exited
+ink_snapclose(Pid) ->
+    try
+        ink_close(Pid)
+    catch
+        exit:{noproc, _CallDetails} ->
+            ok
+    end.
+
 -spec ink_doom(pid()) -> {ok, [{string(), string(), string(), string()}]}.
 %% @doc
 %% Test function used to close a file, and return all file paths (potentially
@@ -789,12 +801,10 @@ handle_cast({maybe_defer_shutdown, ShutdownType, From}, State) ->
     {noreply, State};
 handle_cast({complete_shutdown, ShutdownType, From}, State) ->
     lists:foreach(
-        fun(SnapPid) -> ok = ink_close(SnapPid) end,
-        lists:filter(
-            fun is_process_alive/1,
-            lists:map(
+        fun(SnapPid) -> ok = ink_snapclose(SnapPid) end,
+        lists:map(
                 fun(Snapshot) -> element(1, Snapshot) end,
-                State#state.registered_snapshots))),
+                State#state.registered_snapshots)),
     shutdown_manifest(State#state.manifest),
     case ShutdownType of
         doom ->
@@ -1628,5 +1638,29 @@ loop() ->
         stop ->
             ok
     end.
+
+close_no_crash_test_() ->
+    {timeout, 60, fun close_no_crash_tester/0}.
+
+close_no_crash_tester() ->
+    RootPath = "test/test_area/journal",
+    build_dummy_journal(),
+    CDBopts = #cdb_options{max_size=300000, binary_mode=true},
+    {ok, Inker} =
+        ink_start(
+            #inker_options{
+                root_path=RootPath,
+                cdb_options=CDBopts,
+                compression_method=native,
+                compress_on_receipt=true}),
+
+    SnapOpts =
+        #inker_options{
+            start_snapshot=true, bookies_pid = self(), source_inker=Inker},
+    {ok, InkSnap} = ink_snapstart(SnapOpts),
+
+    exit(InkSnap, kill),
+    ok = ink_close(Inker),
+    clean_testdir(RootPath).
 
 -endif.
--- a/src/leveled_inker.erl
+++ b/src/leveled_inker.erl
@@ -134,6 +134,11 @@
 -define(JOURNAL_FILEX, "cdb").
 -define(PENDING_FILEX, "pnd").
 -define(TEST_KC, {[], infinity}).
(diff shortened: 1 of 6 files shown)


src/leveled_inker.erl
```diff
@@ -778,8 +773,41 @@ handle_cast({remove_logs, ForcedLogs}, State) ->
...
+            % complete_shutdown cast is sent
+            leveled_log:log(i0026, [N]),
+            timer:sleep(?SHUTDOWN_PAUSE)
+    end,
+    gen_server:cast(self(), {complete_shutdown, ShutdownType, From}),
+    {noreply, State};
+handle_cast({complete_shutdown, ShutdownType, From}, State) ->
+    lists:foreach(
+        fun(SnapPid) -> ok = ink_close(SnapPid) end,
+        lists:filter(
```
ThomasArts: Matter of taste, but I find this harder to read than a list comprehension... up to you
In this case, keep the foreach, but filter in the comprehension:
```
[ Snap || {Snap, _TS, _SQN} <- State#state.registered_snapshots, is_process_alive(Snap) ]
```
On a more Meta level... should ink_close(SnapPid) not be able to deal with Pids that are not alive?
After all, there is always the possibility that the snapshot pids terminates just between the check and asking it to close?

What happens now if the pid is when you come to close is still that this may occasionally crash.
But since you know you're shutting down, you could just catch the `ink_close` instead of matching on `ok`. After all, why stop the cleanup if one of the snapshots exists or does not return ok?

src/leveled_inker.erl
```diff
@@ -866,13 +895,6 @@ start_from_file(InkOpts) ->
                     clerk = Clerk}}.
 
 
--spec shutdown_snapshots(list(registered_snapshot())) -> ok.
-%% @doc
-%% Shutdown any snapshots before closing the store
-shutdown_snapshots(Snapshots) ->
-    lists:foreach(fun({Snap, _TS, _SQN}) -> ok = ink_close(Snap) end,
```
ThomasArts, line 873: This original code failed when a snapshot already was closed or did respond with something else than `ok`.
However, in a cleanup, what does it give you when you crash the gen_server loop because you cannot cleanup?

src/leveled_penciller.erl
```diff
@@ -1156,7 +1157,36 @@ handle_cast({remove_logs, ForcedLogs}, State) ->
...
+            % release and have their release messages queued before the
+            % complete_shutdown cast is sent
+            leveled_log:log(p0042, [N]),
+            timer:sleep(?SHUTDOWN_PAUSE)
+    end,
+    gen_server:cast(self(), {complete_shutdown, ShutdownType, From}),
+    {noreply, State};
+handle_cast({complete_shutdown, ShutdownType, From}, State) ->
+    lists:foreach(
+        fun(Snap) -> ok = pcl_close(Snap) end,
```
ThomasArts: symmetric argument to ink_close in inker

src/leveled_pmanifest.erl
```diff
@@ -88,22 +89,23 @@
...
-                    basement :: integer(),
+                    pending_deletes = dict:new() :: dict:dict(), 
+                    basement :: non_neg_integer(),
                         % Currently the lowest level (the largest number)
-                    blooms :: any() % actually a dict but OTP 16 compatability
-                        % A dictionary mapping PIDs to bloom filters
+                    blooms :: dict:dict()
                     }).      
 
+-type fake_pid() :: pid_a1|pid_a2|pid_a3|pid_a4.
```
ThomasArts: is this to please dialyzer because some tests use a symbolic representation of pids?
If so, I were to fix those tests by defining variables:
`PidA1 = spawn(fun() -> ok end)`
and use such fake pids instead. At least they have the right type and they can be compared on equality, but nothing more.

src/leveled_pmanifest.erl
```diff
@@ -538,10 +540,9 @@ mergefile_selector(Manifest, LevelIdx, {grooming, ScoringFun}) ->
 %% be received in parallel to the manifest ebing updated, so the updated
 %% manifest must not trample over any accrued state in the manifest.
 merge_snapshot(PencillerManifest, ClerkManifest) ->
-    ClerkManifest#manifest{snapshots =
-                                PencillerManifest#manifest.snapshots,
-                            min_snapshot_sqn =
-                                PencillerManifest#manifest.min_snapshot_sqn}.
+    ClerkManifest#manifest{
+        snapshots = PencillerManifest#manifest.snapshots,
```
ThomasArts, line 543: Eh... should that not be a list and therefore never match the atom `snapshots`?

If so, this seems untested code

ThomasArts (commented):
Some "up to choice" comments, but have a serious look at pmanifest line 544

src/leveled_pmanifest.erl
martinsumner: It isn't matching.  There are two versions of the manifest, both manifest records.  We want to take everything from the Clerk's manifest going forward (which has just completed a file compaction event), but still use the snapshot information taken from the Pencillers manifest (as it is the Penciller not the Clerk that has been receiving updates to this).  So it takes the Clerks Manifest and just sets the snapshot information on the record to be that of the Penciller.

src/leveled_pmanifest.erl
martinsumner: Fixed as suggested

src/leveled_inker.erl
martinsumner: I've changes to catch the noproc scenario, which removes the need for the filter.

src/leveled_penciller.erl
martinsumner: Changed as with Inker

ThomasArts (approved):

PR #412, opened 2023-09-27
Compile on Erlang 26.1

michaelklishin:


src/leveled_bookie.erl
```diff
@@ -701,7 +701,7 @@ book_returnfolder(Pid, RunnerType) ->
                             {async, Runner::fun(() -> term())}
                                 when Bucket::term(),
                                      StartKey::term(),
-                                     FoldFun::fun((Bucket, Key | {IndexVal, Key}, Acc) -> Acc),
+                                     FoldFun::fun((Bucket, StartKey | {IndexVal, StartKey}, Acc) -> Acc),
```
martinsumner: The FoldFun can take any Key where Key::term() (not necessarily just the StartKey).  So probably need to add a definition of Key rather than change to StartKey.

martinsumner:
Related PR - https://github.com/martinsumner/leveled/pull/408/files

src/leveled_bookie.erl
michaelklishin: OK, thank you for clarifying. It seemed like a typo at first.

michaelklishin:
@martinsumner thank you for the feedback, I have pushed a correction

martinsumner (approved):

Issue #413, opened 2023-10-12
Allow full bookie snapshots that support query API


martinsumner:
The Bookie query API will take the snapshot as part of the API call.

Snapshots of the penciller and inker can be taken separately and returned, should one have knowledge of the internals and want to run repeated queries on the same snapshot.

It would be preferable to have a snapshot that looks like a Bookie, and can support multiple query API calls without needing to refresh the snapshot.

Providing this feature will ease the resolution of https://github.com/nhs-riak/riak_core/issues/1

PR #414, opened 2023-10-13
Allow snapshots to be reused in queries

martinsumner:
Allow for a full bookie snapshot to be re-used for multiple queries, not just KV fetches.

### DIFF
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -1537,28 +1537,40 @@ handle_call(Msg, _From, State) ->
 handle_cast({log_level, LogLevel}, State) ->
     PCL = State#state.penciller,
     INK = State#state.inker,
-    Monitor = element(1, State#state.monitor),
     ok = leveled_penciller:pcl_loglevel(PCL, LogLevel),
     ok = leveled_inker:ink_loglevel(INK, LogLevel),
-    ok = leveled_monitor:log_level(Monitor, LogLevel),
+    case element(1, State#state.monitor) of
+        no_monitor ->
+            ok;
+        Monitor ->
+            leveled_monitor:log_level(Monitor, LogLevel)
+    end,
     ok = leveled_log:set_loglevel(LogLevel),
     {noreply, State};
 handle_cast({add_logs, ForcedLogs}, State) ->
     PCL = State#state.penciller,
     INK = State#state.inker,
-    Monitor = element(1, State#state.monitor),
     ok = leveled_penciller:pcl_addlogs(PCL, ForcedLogs),
     ok = leveled_inker:ink_addlogs(INK, ForcedLogs),
-    ok = leveled_monitor:log_add(Monitor, ForcedLogs),
+    case element(1, State#state.monitor) of
+        no_monitor ->
+            ok;
+        Monitor ->
+            leveled_monitor:log_add(Monitor, ForcedLogs)
+    end,
     ok = leveled_log:add_forcedlogs(ForcedLogs),
     {noreply, State};
 handle_cast({remove_logs, ForcedLogs}, State) ->
     PCL = State#state.penciller,
     INK = State#state.inker,
-    Monitor = element(1, State#state.monitor),
     ok = leveled_penciller:pcl_removelogs(PCL, ForcedLogs),
     ok = leveled_inker:ink_removelogs(INK, ForcedLogs),
-    ok = leveled_monitor:log_remove(Monitor, ForcedLogs),
+    case element(1, State#state.monitor) of
+        no_monitor ->
+            ok;
+        Monitor ->
+            leveled_monitor:log_remove(Monitor, ForcedLogs)
+    end,
     ok = leveled_log:remove_forcedlogs(ForcedLogs),
     {noreply, State}.
 
--- a/src/leveled_bookie.erl
+++ b/src/leveled_bookie.erl
@@ -1259,10 +1259,12 @@ init([Opts]) ->
         {Bookie, undefined} ->
             {ok, Penciller, Inker} = 
                 book_snapshot(Bookie, store, undefined, true),
+            NewETS = ets:new(mem, [ordered_set]),
             leveled_log:log(b0002, [Inker, Penciller]),
             {ok,
                 #state{penciller = Penciller,
                         inker = Inker,
+                        ledger_cache = #ledger_cache{mem = NewETS},
                         is_snapshot = true}}
     end.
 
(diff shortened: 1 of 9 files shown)


martinsumner:
https://github.com/martinsumner/leveled/issues/413

src/leveled_bookie.erl
```diff
@@ -1849,22 +1886,43 @@ set_options(Opts, Monitor) ->
...
+%% runners post-query cleanup action
+%% 
+%% When the bookie is a snapshot, a fresh snapshot should not be taken, the
+%% previous snapshot should be used instead.  Also the snapshot should not be
+%% closed as part of the post-query activity as the snapshot may be reused, and
+%% should be manually closed.
 return_snapfun(State, SnapType, Query, LongRunning, SnapPreFold) ->
-    case SnapPreFold of
-        true ->
+    CloseFun =
```
ThomasArts, line 1912: Sanity check:
Following flow it can be seen that `pcl_close` always returns `ok` if it returns... which means that `ink_close` will be called eventually. But also ink_close will always return `ok` if it returns (there is no corresponding match in line 1918)

Alternative is to check for Pid right away when function is created:
```
fun(LS0, JS0) when is_pid(JS0) ->
    fun() -> ok = leveled_penciller:pcl_close(LS0), leveled_inker:ink_close(JS0) end;
    (LS0, _) ->
       fun() -> leveled_penciller:pcl_close(LS0) end
   end
```

src/leveled_bookie.erl
ThomasArts, line 1: Not completely clear to me whether the extra logging, via `leveled_monitor:log_add(Monitor, ForcedLogs)` is cleaned up when the process dies. There is a DOWN handling that does not call log_remove, but probably that is handled elsewhere anyway.

src/leveled_penciller.erl
```diff
@@ -654,9 +654,12 @@ init([LogOpts, PCLopts]) ->
...
                                                 LongRunning),
             leveled_log:log(p0001, [self()]),
-            {ok, State#state{is_snapshot = true,
-			     bookie_monref = BookieMonitor,
-			     source_penciller = SrcPenciller}};
+            {ok,
+                State#state{
+                    is_snapshot = true,
+                    clerk = undefined,
+			        bookie_monref = BookieMonitor,
```
ThomasArts: Ugly tabs in code
```suggestion
                    bookie_monref = BookieMonitor,
```

src/leveled_penciller.erl
```diff
@@ -654,9 +654,12 @@ init([LogOpts, PCLopts]) ->
...
+			        source_penciller = SrcPenciller}};
```
ThomasArts:
```suggestion
                    source_penciller = SrcPenciller}};
```

src/leveled_penciller.erl
```diff
@@ -1150,22 +1153,33 @@ handle_cast({fetch_levelzero, Slot, ReturnFun}, State) ->
     ReturnFun(lists:nth(Slot, State#state.levelzero_cache)),
     {noreply, State};
 handle_cast({log_level, LogLevel}, State) ->
-    PC = State#state.clerk,
-    ok = leveled_pclerk:clerk_loglevel(PC, LogLevel),
-    ok = leveled_log:set_loglevel(LogLevel),
+    case State#state.clerk of
```
ThomasArts: This code is repeated so often that it almost feels worth to add it to leveled_utils as
```
log_if_pid(undefined, _) -> ok;
log_if_pid(Pid, Fun) when is_pid(Pid) -> Fun().
```
And call it in different forms as:
`log_if_pid(State#state.clerk, fun()  -> leveled_pclerk:clerk_loglevel(PC, LogLevel) end)`

Matter of taste

src/leveled_runner.erl
```diff
@@ -301,13 +275,7 @@ tictactree(SnapFun, {Tag, Bucket, Query}, JournalCheck, TreeSize, Filter) ->
 
             % Close down snapshot when complete so as not to hold removed
             % files open
-            ok = leveled_penciller:pcl_close(LedgerSnap),
-            case JournalCheck of
```
ThomasArts, line 305: Now you build the afterfunction as part of the SnapFun.
But it is hard to follow from the code whether indeed you do this JournalCheck or in line 608 DefferedFetch check.
The code gets clearer, but difficult to validate that it has the same behaviour.
In leveled_bookie the return_snap_fun seems to always close both files, whereas code here suggests that that need not always be the case.

src/leveled_runner.erl
```diff
@@ -602,16 +556,6 @@ foldobjects(SnapFun, Tag, KeyRanges, FoldObjFun, DeferredFetch,
                                                                 LastModRange,
                                                                 LimitByCount)
                 end,
-            AfterFun = 
-                fun() ->
-                    ok = leveled_penciller:pcl_close(LedgerSnapshot),
-                    case DeferredFetch of 
-                        {true, false} ->
```
ThomasArts, line 609: Here we actively did not close the JournalSnapshot.. but is that still the case in the refactored code?

src/leveled_codec.erl
ThomasArts, line 1: This seems rather unrelated to the rest of the refactoring?

martinsumner: This came up when using this patch in riak_kv, it is related to noise reduction in logs.  It isn't normally expected for a tag not to be in the strategy, and so ic012 is `WARN` level log - generally this logs existed because we expect that there may have been some sort of key corruption.  However, when we set a dummy tag because there was a MPUT, that is expected.

The change stops the log noise in this case, but also makes sure via the code coverage checks that we are checking a genuinely invalid tag in the tests

martinsumner: I will add a comment, as it took me a long time to work out why I'd done this, or maybe re-organise the code to make this more obvious

src/leveled_bookie.erl
martinsumner: The monitor and logging is specific to the instance of leveled, and dies with the instance.  The snapshots don't have a monitor process (so stats aren't gathered from snapshots).  the change for this PR is to make sure that when we add_logs to the monitor (to tell it to override the default log level for a given log reference), if we do this to a snapshot it doesn't add to the monitor which doesn't exist.

src/leveled_runner.erl
martinsumner: The tests are comprehensive on this I think, so I'm happy with the risks of the change.  

The only reason for having a JournalSnapshot was to do the JournalCheck.  Previously this was a bit specific ... as a developer I know that for this runner the JournalSnapshot will only present when JournalCheck is true - so test for JournalCheck.  Now with AfterFun/SnapFun it is cleaner that only the Bookie, not the Runner, needs knowledge of whether a snapshot was taken (and therefore if it should be closed).

src/leveled_runner.erl
martinsumner: Yes, as above.  Before we were looking for specific flags to recall if a Journal snapshot would have been taken,  But now, as we create the AfterFun when we create the SnapFun there is no need for function-specific knowledge downstream.

src/leveled_bookie.erl
martinsumner: I think I'm happy with it as-is.  Double-checked as well `ok` is the only return.

src/leveled_penciller.erl
martinsumner: Fixed

src/leveled_penciller.erl
martinsumner: Fixed

src/leveled_penciller.erl
martinsumner: Repeated code now in function

martinsumner:
Reviewed https://github.com/martinsumner/leveled/pull/415

PR #415, opened 2023-10-17
Mas d31 i413

martinsumner:
Improve usability of snapshots.

Full snapshots of a store would only work for basic GET requests previously.  No support for the query API f for head-only mode.

This allows a snapshot bookie to be taken which works as a normal Bookie, for all read-only operations.


ThomasArts (commented):
pretty much same "remarks" as for 414.
But now I was able to run tests and compile on OTP25 and OTP26
(warning for fqc in OTP26. I guess that package needs some update, but nothing problematic)

ThomasArts (approved):

PR #417, opened 2023-10-25
Add compression controls

martinsumner:
Add configuration options to allow for a compression algorithm of `none` to disable compression altogether.  Also an option to change the point in the LSM tree when compression is applied.


priv/leveled_multi.schema
```diff
@@ -39,8 +39,8 @@
 %% Can be lz4 or native (which will use the Erlang native zlib compression)
 %% within term_to_binary
 {mapping, "multi_backend.$name.leveled.compression_method", "riak_kv.multi_backend", [
-  {default, lz4},
-  {datatype, atom},
+  {datatype, {enum, [native, lz4, none]}},
+  {default, native}
```
ThomasArts, line 43: Sure to change the default here?

include/leveled.hrl
```diff
@@ -55,6 +55,7 @@
 -record(sst_options,
                         {press_method = native
                             :: leveled_sst:press_method(),
+                        press_level = 1 :: non_neg_integer(),
```
ThomasArts: In leveled_bookie you define COMPRESSION_LEVEL to be 1.
I think, it would be nicer to define that default here and use it:
```suggestion
                        press_level = ?COMPRESSION_LEVEL :: non_neg_integer(),
```
Otherwise you encode the default in more than one place and that might be leading to future mistakes

src/leveled_sst.erl
```diff
@@ -78,6 +78,10 @@
 -define(NOLOOK_SLOTSIZE, 256).
 -define(NOLOOK_BLOCKSIZE, {56, 32}). % 4x + y = ?NOLOOK_SLOTSIZE
 -define(COMPRESSION_LEVEL, 1).
+    % When using native compression - how hard should the compression code
+    % try to reduce the size of the compressed output. 1 Is to imply minimal
+    % effort, 6 is default:
```
ThomasArts: but not the default of leveled, because here 1 is default

ThomasArts (approved):
Looks good to me. Just double check you want to chaneg the default from lz4 to native at indicated place

priv/leveled_multi.schema
martinsumner: It was a mistake.  The default for leveled backend was changed to native, but in that fix, forgot to change the multi-backend.  So this realigns the two again.

For riak, as it supports a HTTP API, there tends to be a lot of stuff that ends up being stored after base64 encoding.  In this case we get a 30% compression improvement native via lz4. The throughput improvement we've seen with lz4 due to efficient CPU has always been no more than about 3% except for contrived situations.

include/leveled.hrl
martinsumner: I agree.  There's quite a few other things where we end up with double-definitions of defaults, but I think it makes sense to correct them all as part of this commit.  In general all configurable items should have their defaults in `leveled.hrl`

src/leveled_sst.erl
martinsumner: Changed to COMPRESSION_FACTOR as this is different to the configurable COMPRESSION_LEVEL.  Will add reference to default being OTP not leveled.

PR #423, opened 2023-11-23
Mas d31 nhskv16

martinsumner:
Add perf_SUITE test to measure guess/estimate accuracy and speed for counts.

Add improvement to `recovery_SUITE/replace_everything/1` test to handle intermittent failures.

### DIFF
--- a/src/leveled_imanifest.erl
+++ b/src/leveled_imanifest.erl
@@ -19,7 +19,8 @@
         reader/2,
         writer/3,
         printer/1,
-        complete_filex/0
+        complete_filex/0,
+        get_cdbpids/1
         ]).         
 
 -define(MANIFEST_FILEX, "man").
@@ -27,7 +28,7 @@
 -define(SKIP_WIDTH, 16).
 -define(MANIFESTS_TO_RETAIN, 5).
 
--type manifest() :: list({integer(), list()}).
+-type manifest() :: list({integer(), list(manifest_entry())}).
 %% The manifest is divided into blocks by sequence number, with each block
 %% being a list of manifest entries for that SQN range.
 -type manifest_entry() :: {integer(), string(), pid()|string(), any()}.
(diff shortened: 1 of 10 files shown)


martinsumner:
https://github.com/nhs-riak/riak_kv/issues/16

martinsumner:
Results from running perf_SUITE test with different sizes and OTP versions:

Screenshot 2023-11-23 at 09 26 57

Screenshot 2023-11-23 at 09 27 47

martinsumner:
The initial testing of the `guess` method for object count estimation indicated it was surprisingly slow.  Counting 1:1000 keys took 10% of the time - although there is additional work necessary to skip using segment lists, this  intuitively appears to be out by an order of magnitude.

Profiling indicated that the majority of time was being spent in two places:

- within `expand_list_by_pointer/5`, in particular though the use of `++`;
- in the `find_pos/4` function, in particular repeated calls to `lists:member/2`.

the expand_list_by_pointers has been refactored and simplified.  The member check under find_pos can now be avoided where SegmentIds are not randomly distributed.

This has reduced the time to `guess` object counts by 65% in OTP 22 (and by 75% by also changing to OTP 26).

The outstanding problem, is that `find_pos` is still perhaps unexpectedly dominant in CPU time:

```
leveled_sst:'-segment_checker/1-fun-2-'/5                   400000000    19.31  15560374  [      0.04]

leveled_sst:find_pos/4                                      403085303    35.05  28242226  [      0.07]
```

Further, there appears to be a discrepancy between performance in eunit and ct conditions of this function.

martinsumner:
Relative load performance with this PR:

Screenshot 2023-11-24 at 22 11 14

martinsumner:
Relative fold_heads performance with this PR:

Screenshot 2023-11-24 at 22 11 29

PR #427, opened 2023-12-14
Mas d31 nhskv16sstpclv2

martinsumner:
Further improvements to `leveled_sst`, to:
- reduce the use of `++` or `lists:append/2` where a BigList is prepended to a ShortList;
- reduce the duplication of functions, where the same thing is done differently in different paths through the code;
- be more specific about what blocks need to be deserialised when using a key range, and trimming prior to accumulation.

### DIFF
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -79,8 +79,6 @@
 -define(DELETE_TIMEOUT, 10000).
 -define(TREE_TYPE, idxt).
 -define(TREE_SIZE, 16).
--define(TIMING_SAMPLECOUNTDOWN, 20000).
--define(TIMING_SAMPLESIZE, 100).
 -define(BLOCK_LENGTHS_LENGTH, 20).
 -define(LMD_LENGTH, 4).
 -define(FLIPPER32, 4294967295).
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -378,14 +378,10 @@ sst_newlevelzero(
                          infinity),
     ok =
         case Fetcher of
-            FetchSlots when is_list(Fetcher) ->
-               gen_statem:cast(Pid, {complete_l0startup, FetchSlots});
-            _ ->
-                % Fetcher is a function
-                gen_statem:cast(Pid, {sst_returnslot, none, Fetcher, Slots})
-                % Start the fetch loop (async).  Having the fetch loop running
-                % on async message passing means that the SST file can now be
-                % closed while the fetch loop is still completing
+            SlotList when is_list(SlotList) ->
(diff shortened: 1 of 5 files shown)


martinsumner:
Comparison shows improvement sin LoadTime.  

P427 vs P426 vs D31

Degradation in GetTime when testing in OTP 26 is believed to be local machine issues with page cache.

src/leveled_sst.erl
```diff
@@ -1450,85 +1368,65 @@ fetch(LedgerKey, Hash,
...
+        N ->
+            {LSlot, MidSlots, RSlot} =
+                case N of
+                    2 ->
+                        [Slot1, Slot2] = Slots,
+                        {Slot1, [], Slot2};
+                    N ->
+                        [Slot1|_Rest] = Slots,
+                        SlotN = lists:last(Slots),
+                        {Slot1, lists:sublist(Slots, 2, N - 2), SlotN}
```
ThomasArts: For the record. I tried:
```
   [S1 | Rest] = Slots,
   {S1, lists:droplast(Rest), lists:last(Rest)}
```
but my intuition was wrong, it is sometimes faster, sometimes not.

ThomasArts: Advantage, one does not need a separate case for 2 elements.
but you don't need that case if you do:
```
   [Slot1|Rest] = Slots,
   SlotN = lists:last(Slots),
    {Slot1, lists:sublist(Rest, 1, N - 1), SlotN}
    %% alternative lists:sublist(Slots, 2, N - 2),

24> lists:sublist([1,2], 2, 0).
[]
```

src/leveled_sst.erl
```diff
@@ -2560,66 +2424,141 @@ binaryslot_trimmedlist(FullBin, StartKey, EndKey,
...
+            lt_mid;
         {false, true} ->
-            [Block1, Block2, MidBlockList];
+            le_mid;
         {false, false} ->
-            [Block1, Block2, MidBlockList, Block4, Block5]
+            all_blocks
     end.
 
+previous_block_required({not_present, not_present}, _SK) ->
```
ThomasArts, line 2527: this clause is strictly spoken not needed

ThomasArts (approved):

src/leveled_sst.erl
martinsumner: The function is not called super-frequently, so I don't think the extra cause is harmful ... save a future reviewer having to double-check the impact of erlang term order

martinsumner:
Chart showing relative performance between develop-3.1 and PR427 (OTP22 and OTP 26):

Testv5 - Comparison of PR427

PR #434, opened 2024-04-12
Extend perf_SUITE

martinsumner:
This is v6 of the perf_SUITE tests.  The test adds a complex index entry to every object, and then adds a new test phase to test regex queries.

There are three profiles added so the full, mini and profiling versions of perf_SUITE can be run without having to edit the file itself:

e.g. ./rebar3 as perf_mini do ct --suite=test/end_to_end/perf_SUITE

When testing as `perf_prof` summarised versions of the eprof results are now printed to screen.

The volume of keys within the full test suite has been dropped ... just to make life easier so that test run times are not excessively increase by the new features.

### DIFF
--- a/test/end_to_end/testutil.erl
+++ b/test/end_to_end/testutil.erl
@@ -68,7 +68,7 @@
             compact_and_wait/1]).
 
 -define(RETURN_TERMS, {true, undefined}).
--define(SLOWOFFER_DELAY, 10).
+-define(SLOWOFFER_DELAY, 40).
 -define(V1_VERS, 1).
 -define(MAGIC, 53). % riak_kv -> riak_object
 -define(MD_VTAG,     <<"X-Riak-VTag">>).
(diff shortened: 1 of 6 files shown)


test/end_to_end/testutil.erl
```diff
@@ -809,26 +813,42 @@ put_altered_indexed_objects(Book, Bucket, KSpecL, RemoveOld2i, V) ->
...
+                        case book_riakput(Book, O, DeltaSpecs) of
+                            ok -> ok;
+                            pause -> timer:sleep(?SLOWOFFER_DELAY)
+                        end,
+                    ThisProcess ! {R, DeltaSpecs}
+                end,
+            spawn(PutFun),
+            AccOut =
+                receive
+                    {ok, NewSpecs} -> Acc;
```
ThomasArts, line 843: timer:sleep always returns `ok` Thus R is never `pause` in line 830

ThomasArts (commented):
Proposal for profiles in separate PR (437), otherwise fine.

ThomasArts (approved):

PR #435, opened 2024-04-12
Switch to re2 library

martinsumner:
The re2 library, though not offering the complete functionality of pcre, is known to be much faster in many scenarios.

Switch to re2 in preparation for the OTP switch to re2 in OTP28

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ cover
 cover_*
 .eqc-info
 leveled_data/*
+compile_commands.json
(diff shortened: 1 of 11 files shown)


martinsumner:
There are some issues with using re2 in riak.  It is not possible to pass/use a reference to a compiled regular expression between nodes.  Further there are significant performance benefits to using [filter expressions](https://github.com/martinsumner/leveled/tree/mas-d31-mas.i1433-filterexpression) rather than regular expression.  This is a better way to improve filter speed.

So the intention is now to wait for the OTP-implemented version to be available.

PR #438, opened 2024-04-27
Mas d31 i433re2capture

martinsumner:
Allow for evaluation and filtering of 2i terms.

Eval: Have a language of functions whereby, a map of Key/Value pairs can be extracted by applying those functions in a pipeline.

Filter: Have a language of functions and operators to apply a logic expression to determine if a term matches a filter, based on the entries in the map output from the eval pipeline.

The aim is to have an alternative to regular expression filtering in Riak.  Sometimes the auto-generation of regular expressions become complex, and somethings (such as date ranges) and potentially impossible.  The aim of the filter expression is to do this in a more natural language.

### DIFF
--- a/src/leveled_codec.erl
+++ b/src/leveled_codec.erl
@@ -114,9 +114,9 @@
         lookup|no_lookup.
 -type actual_regex() ::
         reference()|{re_pattern, term(), term(), term(), term()}.
--type capture_filter_fun() :: fun((list({binary(), binary()})) -> boolean()).
+-type capture_filter_fun() :: fun((#{atom() => string()}) -> boolean()).
 -type capture_reference() :: 
-        {capture, actual_regex(), list(binary()), capture_filter_fun()}.
+        {capture, actual_regex(), list(atom()), capture_filter_fun()}.
 -type regular_expression() ::
         actual_regex()|undefined|capture_reference().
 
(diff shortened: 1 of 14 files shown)


src/leveled_codec.erl
```diff
@@ -112,10 +112,17 @@
         {index_specs(), infinity|integer()}. % {KeyChanges, TTL}
 -type maybe_lookup() ::
         lookup|no_lookup.
+-type actual_regex() ::
+        reference()|{re_pattern, term(), term(), term(), term()}.
```
ThomasArts, line 116: is there a reason not to use the opaque type `re:mp()` instead of `re_patten...`?

src/leveled_eval.erl
```diff
@@ -0,0 +1,560 @@
...
+apply_eval({eval, Eval}, Term, Key, AttrMap) ->
+    apply_eval(Eval, Term, Key, AttrMap);
+apply_eval({'PIPE', Eval1, 'INTO', Eval2}, Term, Key, AttrMap) ->
+    apply_eval(Eval2, Term, Key, apply_eval(Eval1, Term, Key, AttrMap));
+apply_eval({
+        delim, {identifier, _, InKey}, {string, _, Delim}, ExpKeys},
+        Term, Key, AttrMap) ->
+    TermToSplit = term_to_process(InKey, Term, Key, AttrMap),
+    CptTerms = string:split(TermToSplit, Delim, all),
+    L = min(length(CptTerms), length(ExpKeys)),
```
ThomasArts, line 39: in OTP 26 there is now a `How` parameter for zip. Using `trim` would reduce this code. But only introduced for OTP26.

ThomasArts (approved):

src/leveled_codec.erl
martinsumner: [Not exported until > OTP26](https://github.com/erlang/otp/commit/eb4a9c1e45bccf8d5463ec69198db41149eda823)

martinsumner:
Replaced with https://github.com/martinsumner/leveled/pull/440

PR #439, opened 2024-05-07
Add support for combination queries on same snapshot point

martinsumner:
Allow for return_keys of logical combinations of queries

### DIFF
--- a/src/leveled_setop.erl
+++ b/src/leveled_setop.erl
@@ -1,0 +1,110 @@
+%% -------- Set Operations ---------
+%%
+%% Support for set operations (i.e on sets of keys) within leveled 
+%%
+
+-module(leveled_setop).
+
+-export([generate_setop_function/1]).
+
+
+%%%============================================================================
+%%% External API
+%%%============================================================================
+
+-spec generate_setop_function(
+        string()) -> fun((list(sets:set(binary()))) -> sets:set(binary())).
+generate_setop_function(EvalString) ->
+    {ok, ParsedEval} = generate_setop_expression(EvalString),
+    fun(ListOfSets) ->
+        apply_setop(ParsedEval, ListOfSets)
+    end.
+
+%%%============================================================================
+%%% Internal functions
+%%%============================================================================
+
+generate_setop_expression(EvalString) ->
+    {ok, Tokens, _EndLine} = leveled_setoplexer:string(EvalString),
+    leveled_setopparser:parse(Tokens).
+
+apply_setop({setop, SetOp}, SetList) ->
+    apply_setop(SetOp, SetList);
+apply_setop(
+        {SetFunctionName, {set_id, _, SetIDa}, {set_id, _, SetIDb}},
+        SetList) ->
+    SetFunction = set_function(SetFunctionName),
+    SetFunction(lists:nth(SetIDa, SetList), lists:nth(SetIDb, SetList));
+apply_setop(
+        {SetFunctionName, {set_id, _, SetIDa}, Condition},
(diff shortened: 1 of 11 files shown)


src/leveled_setoplexer.xrl
```diff
@@ -0,0 +1,24 @@
...
+Rules.
+
+{WhiteSpace} : skip_token.
+
+\(        : {token, {'(', TokenLine}}.
+\)        : {token, {')', TokenLine}}.
+
+OR        : {token, {'UNION', TokenLine}}.
+AND       : {token, {'INTERSECT', TokenLine}}.
+NOT       : {token, {'SUBTRACT', TokenLine}}.
```
ThomasArts, line 13: Kind, of, NOT is in fact a unary operation and SUBTRACT a binary.

This might cause future confusion. If you write NOT $d, then you mean to say the universe apart from whatever is in $d. But the universe is not part of your language.

- But more prominently, if I write `$a SUBTRACT NOT $b` you fail, whereas in set theory you would take the complenet from B and subtract that from A.

It will be hard to implement complement. So not sure you want to add NOT. Also because NOT is very hard to do in regular expressions.

src/leveled_setopparser.yrl
ThomasArts, line 1: In many applications this parser would have a "normalize" function that applies "De Morgan" laws to normalize the SET. This may simplify the final regular expression building, but may be considered out of scope here.

src/leveled_setoplexer.xrl
```diff
@@ -0,0 +1,24 @@
...
+UNION     : {token, {'UNION', TokenLine}}.
+INTERSECT : {token, {'INTERSECT', TokenLine}}.
+SUBTRACT  : {token, {'SUBTRACT', TokenLine}}.
+
+\$[1-8]+     : {token, {set_id, TokenLine, strip_identifier(TokenChars)}}.
```
ThomasArts, line 18: This feels limiting... zero and nine cannot be in the `set_id`? Do you actually mean a number not starting with zero?
`\$[1-9][0-9]*`

ThomasArts: In particular since you take the "nth" set_id from a list, missing the "9" is confusing if you allow "11". 
I like the map you use in filters better, then it does not really matter what representation you use. Is that possible here as well, to provide `apply_setop` with a map of `set_id` to the actual set?

src/leveled_setop.erl
```diff
@@ -0,0 +1,110 @@
...
+%%% Internal functions
+%%%============================================================================
+
+generate_setop_expression(EvalString) ->
+    {ok, Tokens, _EndLine} = leveled_setoplexer:string(EvalString),
+    leveled_setopparser:parse(Tokens).
+
+apply_setop({setop, SetOp}, SetList) ->
+    apply_setop(SetOp, SetList);
+apply_setop(
```
ThomasArts, line 33: base case missing:
```
apply_setop({set_id, _, SetID}, SetList) ->
      lists:nth(SetID, SetList)
 ```
 This follows from 
 ```
 9> F =  leveled_setop:generate_setop_function("$1").
#Fun<leveled_setop.0.97347034>
10> F([[]]).
** exception error: no function clause matching leveled_setop:set_function(set_id) (/Users/thomas/Quviq/Customers/NHS/leveled/src/leveled_setop.erl, line 54)
     in function  leveled_setop:apply_setop/2 (/Users/thomas/Quviq/Customers/NHS/leveled/src/leveled_setop.erl, line 49)

 ```

PR #440, opened 2024-05-17
Filter Expression support

martinsumner:
Allow indexes to be evaluated into a Key/Value map using an eval pipeline, and then filtered using a filter expression


rebar.config
```diff
@@ -4,11 +4,11 @@
 
 {xref_checks,
   [undefined_function_calls,undefined_functions,
-    locals_not_used,
     deprecated_function_calls, deprecated_functions]}.
 
 {cover_excl_mods,
-  [testutil,
+  [leveled_filterlexer, leveled_filterparser, leveled_evallexer, leveled_evalparser,
```
ThomasArts: What is the reason to exclude parser and lexer from coverage?
When I ran quickcheck property, I cover all code apart from some error cases. Was adding needless negative tests the reason?

src/leveled_codec.erl
```diff
@@ -112,10 +112,17 @@
         {index_specs(), infinity|integer()}. % {KeyChanges, TTL}
 -type maybe_lookup() ::
         lookup|no_lookup.
+-type actual_regex() ::
+        {re_pattern, term(), term(), term(), term()}.
```
ThomasArts, line 134: would it not be nicer to use `re:mp()` here, to hide the details?

src/leveled_eval.erl
```diff
@@ -0,0 +1,648 @@
...
+-module(leveled_eval).
+
+-export([apply_eval/4, generate_eval_expression/2, generate_eval_function/2]).
+
+%%%============================================================================
+%%% External API
+%%%============================================================================
+
+generate_eval_function(EvalString, Substitutions) ->
+    {ok, ParsedEval} = generate_eval_expression(EvalString, Substitutions),
```
ThomasArts, line 15:
```suggestion
    CodePointList = unicode:characters_to_list(EvalString),
    {ok, ParsedEval} = generate_eval_expression(CodePointList, Substitutions),
```

src/leveled_eval.erl
```diff
@@ -0,0 +1,648 @@
...
+    fun(Term, Key) ->
+        apply_eval(ParsedEval, Term, Key, maps:new())
+    end.
+
+generate_eval_expression(EvalString, Substitutions) ->
+    {ok, Tokens, _EndLine} = leveled_evallexer:string(EvalString),
```
ThomasArts:
```suggestion
    CodePointList = unicode:characters_to_list(EvalString),
    {ok, Tokens, _EndLine} = leveled_evallexer:string(CodePointList),
```

src/leveled_eval.erl
```diff
@@ -0,0 +1,648 @@
...
+    ?assertNot(maps:is_key(<<"doh">>, EvalOut9)),
+
+    %% Age at 30 April 2024
+    EvalString10 =
+        EvalString5 ++ 
+        " | index($dob, 4, 4, $birthday)"
+        " | map($birthday, <=, ((\"0430\", 2024)), 2023, $yoc)"
+        " | subtract($yoc, $yob, $age)"
+        " | add($age, 1, $age_next)"
+        " | to_string($age, $age)"
```
ThomasArts, line 579: Didn't realize one can overload variables to have different types... probably not the nicest thing to use in an example.

src/leveled_evallexer.xrl
```diff
@@ -0,0 +1,55 @@
+%% Lexer for eval expressions
+
+Definitions.
+WhiteSpace  = ([\t\f\v\r\n\s]+)
```
ThomasArts, line 4: When reading the documentation, I would say `\s` subsumes the others.

src/leveled_evallexer.xrl
```diff
@@ -0,0 +1,55 @@
...
+to_string    : {token, {to_string, TokenLine}}.
+add          : {token, {add, TokenLine}}.
+subtract     : {token, {subtract, TokenLine}}.
+map          : {token, {map, TokenLine}}.
+pcre         : {token, {pcre, TokenLine}}.
+
+=       : {token, {comparator, '=', TokenLine}}.
+<       : {token, {comparator, '<', TokenLine}}.
+>       : {token, {comparator, '>', TokenLine}}.
+<>      : {token, {comparator, '<>', TokenLine}}.
```
ThomasArts:
```suggestion
```

src/leveled_codec.erl
martinsumner: The type is not exported until OTP 27 I think - so you can't refer to it.

rebar.config
martinsumner: It is the auto-generated files that were the problem.  The eunit/ct tests cover leveled_filter, and leveled_eval modules - but there was a large number of coverage holes in the auto-generated files.

Seen as they are auto-generated, I din't think they were relevant to coverage.

src/leveled_codec.erl
ThomasArts: in leveled_runner you replaced it by `any()`, but I prefer here to wait until OTP27 is lowest supported and replace it then.

ThomasArts (approved):

ThomasArts (approved):

PR #442, opened 2024-07-11
Switch to logger

martinsumner:
Use logger rather than io:format when logging.  The ct tests have besn switched to log to file, testutil/init_per_suite/1 may offer useful guidance on configuring logger with leveled.

As all logs are produced by the leveled_log module, the MFA metadata is uninteresting for log outputs, but can be used for explicit filter controls for leveled logs.


src/leveled_log.erl
```diff
@@ -458,30 +462,11 @@ log_randomtimer(LogReference, Subs, StartTime, RandomProb) ->
...
-    [$0, $0+I];
-i2l(I) ->
-    integer_to_list(I).
-
--spec i3l(non_neg_integer()) -> list().
-i3l(I) when I < 100 ->
-    [$0 | i2l(I)];
-i3l(I) ->
-    integer_to_list(I).
+-spec log_prefix(atom(), non_neg_integer(), pid()) -> io_lib:chars().
```
ThomasArts, line 465: Creating a list for strings is perfectly fine. Just want to point out that using the template functionality build in in `logger` is an alternative.

https://www.erlang.org/doc/apps/kernel/logger_formatter.html#t:template/0

test/end_to_end/appdefined_SUITE.erl
```diff
@@ -1,7 +1,6 @@
 -module(appdefined_SUITE).
--include_lib("common_test/include/ct.hrl").
```
ThomasArts, line 2: I am surprised that this include is not needed??

test/end_to_end/tictac_SUITE.erl
```diff
@@ -16,11 +15,19 @@ all() -> [
...
 
+init_per_suite(Config) ->
+    testutil:init_per_suite([{suite, "tictac"}|Config]),
+    Config.
+
+end_per_suite(Config) ->
+    testutil:end_per_suite(Config).
+
 many_put_compare(_Config) ->
+
```
ThomasArts, line 29:
```suggestion
```

ThomasArts (approved):

test/end_to_end/appdefined_SUITE.erl
martinsumner: So was I, but I don't really use any of the inbuilt ct functionality - there might be a long term task to use it more in the future.

PR #447, opened 2024-08-16
Mas d34 leveled.i446 readcache

martinsumner:
Implement a centralised and configurable read-cache for the ledger, rather than a per-sst file read cache


martinsumner:
The code seems simpler, but there were no performance benefits.  Checking in real-world systems, the distributed cache achieves a high hit ratio, so there are risks with implementing the central cache.

Despite the overall cache size being smaller, the memory footprint was little changed.

Issue #449, opened 2024-09-03
Shutdown and compaction race


martinsumner:
If while a clerk is completing a compaction (i.e. post scoring) an inker is shutdown.  The following may occur.

1. The inker starts the close process, it waits for the iclerk to respond to `leveled_iclerk:clerk_stop(State#state.clerk)`
2. The iclerk responds when it has cast to the inker `leveled_inker:ink_clerkcomplete(State#state.inker, ManifestSlice, FilesToDelete)`.
3. The inker receives the `clerk_complete` message (prior to the `maybe_defer_shutdown`, updates the manifest and casts a request to `leveled_iclerk:clerk_promptdeletions(State#state.clerk, NewManifestSQN, FilesToDelete)` ... but this cast will never be received as the clerk is shutdown.
4. The inker now receives `maybe_defer_shutdown` and then `complete_shutdown` and shuts down all files within the manifest ... but this doesn't include the files which were to be prompted for deletion.
5. As the replaced CDB files are now neither shutdown due to being in the manifest, nor shutdown by delete_pending state timeout - so they are hanging PIDs.
6. when node termination eventually closes the PIDs, they will not clear the files from disk, as they have not been moved to delete_pending.

This is (rarely) detected by `leveled_statemeqc`

PR #450, opened 2024-09-03
Mas i449 directpromptofdeletions

martinsumner:
Previously the inker casts to the clerk so that the clerk can cast about pending deletes to cdb files removed in compactions.  Now the inker casts directly, which addresses some potential race conditions when closes occur at the end of compaction jobs.


martinsumner:
https://github.com/martinsumner/leveled/issues/449

ThomasArts (approved):
Nice!

PR #451, opened 2024-09-17
Make tree compatible with binary L1

martinsumner:
The old leveled_tictac had a pure binary L1.  this was slower than the new map version.

However, in a Riak cluster, when running a merge_tree_range during a rolling update, the fold the query coordinator will initiate a tree.  If this tree is not a map-based tree (as that node has not yet been upgraded), then a node that has been upgraded would previously fail the query as it cannot handle a level 1 in a binary form.  This now enables updated nodes to handle both forms of trees.

Obviously, if the coordinating node has been updated non-updated nodes will crash queries as they cannot handle the tree with the map at Level 1.  The aim is to make it configurable to force non-map trees in a cluster, until all nodes have been upgraded.  So as long as each node understands how to update both non-map trees and map-based trees - evrything should be OK.

### DIFF
--- a/src/leveled_tictac.erl
+++ b/src/leveled_tictac.erl
@@ -54,6 +54,7 @@
 -export([
             new_tree/1,
             new_tree/2,
+            new_tree/3,
             add_kv/4,
             add_kv/5,
             alter_segment/3,
@@ -103,7 +104,7 @@
                         level2 :: array:array()
                         }).
 
--type level1_map() :: #{non_neg_integer() => binary()}.
+-type level1_map() :: #{non_neg_integer() => binary()}|binary().
 
 -type tictactree() ::
     #tictactree{}.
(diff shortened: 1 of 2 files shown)


src/leveled_tictac.erl
```diff
@@ -502,48 +516,71 @@ to_level1_map_loop(<<>>, L1MapAcc, _Idx) ->
...
+    Level1Slice = Level1Pos div 16,
+    HashIntLength = ?HASH_SIZE * 8,
+    Level1BytePos = ?HASH_SIZE * (Level1Pos rem 16),
+    <<PreL1:Level1BytePos/binary,
+        SegLeaf1:HashIntLength/integer,
+        PostL1/binary>> = maps:get(Level1Slice, Level1M),
+    {PreL1, SegLeaf1, PostL1, Level1BytePos};
+extract_level1_slice(Level1B, Level1Pos) when is_binary(Level1B) ->
+    HashIntLength = ?HASH_SIZE * 8,
+    Level1BytePos = ?HASH_SIZE * Level1Pos,
```
ThomasArts, line 571: Why the difference with the original code where you do:
`Level1BytePos = ?HASH_SIZE * (Level1Pos rem 16)`

Ah, I see... the original code has this, then you changed it to `rem 16` for the newer version, but the old version sticks with the value you have here.

ThomasArts (approved):

PR #452, opened 2024-09-18
Make tree compatible with binary L1

martinsumner:
The old leveled_tictac had a pure binary L1.  this was slower than the new map version.

However, in a Riak cluster, when running a merge_tree_range during a rolling update, the fold the query coordinator will initiate a tree.  If this tree is not a map-based tree (as that node has not yet been upgraded), then a node that has been upgraded would previously fail the query as it cannot handle a level 1 in a binary form.  This now enables updated nodes to handle both forms of trees.

Obviously, if the coordinating node has been updated non-updated nodes will crash queries as they cannot handle the tree with the map at Level 1.  The aim is to make it configurable to force non-map trees in a cluster, until all nodes have been upgraded.  So as long as each node understands how to update both non-map trees and map-based trees - evrything should be OK.

### DIFF
--- a/src/leveled_tictac.erl
+++ b/src/leveled_tictac.erl
@@ -55,6 +55,7 @@
 -export([
             new_tree/1,
             new_tree/2,
+            new_tree/3,
             add_kv/4,
             add_kv/5,
             alter_segment/3,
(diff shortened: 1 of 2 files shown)


martinsumner:
Reviewed - https://github.com/martinsumner/leveled/pull/451

PR #454, opened 2024-10-11
Mas d34 i453 eqwalizer

martinsumner:
A large set of changes to make leveled clean for eqwalizer.

The most common issues were:
- a need to clarify types in guards i.e. where a type is something() | undefined - clarify that it cannot be undefined in specific clauses.  The ?IF_DEF macro has been added to check something is not undefined.
- types that were unclear and overlapping, especially the type ledger_key() was over-simplistic, and clarity could be gained by considering primary_key() or query_key() as more specific cases.
- use of invalid types in unit tests, or inaccurate types to support loosely defined unit tests - this was especially true with using string() not binary() for key components.
- there were some dual-use functions, which to clarify types have been split into two single-use functions.


include/leveled.hrl
```diff
@@ -80,9 +89,9 @@
                         timestamp :: integer()}).                      
 
 -record(manifest_entry,
-                        {start_key :: tuple() | undefined,
-                        end_key :: tuple() | undefined,
-                        owner :: pid()|list(),
+                        {start_key :: tuple(),
+                        end_key :: tuple(),
+                        owner :: pid(),
```
ThomasArts: I guess the type system helps here and an owner is never of type  `list()`. But in penciller there is a place where is is assigned the states `levelzero_constructor` which might in theory be `undefined`. So I wonder whether it could be undefined at all... but since none of the type systems have complained about it, the answer must be that this is not possible.

src/leveled_tictac.erl
```diff
@@ -229,12 +236,14 @@ import_tree(ExportedTree) ->
...
     add_kv(TicTacTree, Key, Value, BinExtractFun, false).
 
--spec add_kv(
-    tictactree(), term(), term(), bin_extract_fun(), boolean())
-        -> tictactree()|{tictactree(), integer()}.
+-spec add_kv
+    (tictactree(), term(), term(), bin_extract_fun(), true) ->
+        {tictactree(), integer()};
+    (tictactree(), term(), term(), bin_extract_fun(), false) ->
+        tictactree().
```
ThomasArts, line 243: Ah, Eqwalizer supports dependent types, splendid.

src/leveled_tictac.erl
```diff
@@ -249,12 +258,9 @@ add_kv(TicTacTree, Key, Value, BinExtractFun, ReturnSegment) ->
...
         ),
-    case ReturnSegment of
-        true -> 
-            {UpdatedTree, Segment};
-        false ->
-            UpdatedTree
-    end.
+    {UpdatedTree, Segment};
+add_kv(TicTacTree, Key, Value, BinExtractFun, false) ->
+    element(1, add_kv(TicTacTree, Key, Value, BinExtractFun, true)).
```
ThomasArts, line 263: Not winning the beauty contest. Not sure how to do it better, though.

src/leveled_imanifest.erl
```diff
@@ -28,9 +28,7 @@
 -type manifest() :: list({integer(), list()}).
 %% The manifest is divided into blocks by sequence number, with each block
 %% being a list of manifest entries for that SQN range.
--type manifest_entry() :: {integer(), string(), pid()|string(), any()}.
-%% The Entry should have a pid() as the third element, but a string() may be
-%% used in unit tests
+-type manifest_entry() :: {integer(), string(), pid(), any()}.
```
ThomasArts, line 31: a bit confusing that there also is a type `manifest_entry` in leveled.hrl where it's a record... code would be clearer if that's the same record used here as well?

src/leveled_monitor.erl
```diff
@@ -173,8 +173,10 @@
 
 -spec monitor_start(pos_integer(), list(log_type())) -> {ok, pid()}.
 monitor_start(LogFreq, LogOrder) ->
-    gen_server:start_link(
-        ?MODULE, [leveled_log:get_opts(), LogFreq, LogOrder], []).
+    {ok, Monitor} =
+        gen_server:start_link(
+            ?MODULE, [leveled_log:get_opts(), LogFreq, LogOrder], []),
+    {ok, Monitor}.
```
ThomasArts, line 179: Nice! This actually gets better code. Since the `gen_server:start`  could have returned an error, which would go unnoticed, you now crash if monitoring is not possible.

src/leveled_pclerk.erl
```diff
@@ -439,38 +445,22 @@ grooming_score_test() ->
...
-                                    3,
-                                    999999,
-                                    #sst_options{},
-                                    true,
-                                    false),
-    
-    ME1 = #manifest_entry{owner=PidL3_1},
-    ME1B = #manifest_entry{owner=PidL3_1B},
-    ME2 = #manifest_entry{owner=PidL3_2},
-    ME2NC = #manifest_entry{owner=PidL3_2NC},
```
ThomasArts, line 459: You silently remove some testing here. No longer a relevant test?

src/leveled_penciller.erl
```diff
@@ -291,9 +302,9 @@
                     leveled_codec:ledger_kv()|not_present).
 -type levelzero_returnfun() :: fun((levelzero_cacheentry()) -> ok).
 -type pclacc_fun() ::
-        fun((leveled_codec:ledger_key(),
+        fun((leveled_codec:object_key(),
                 leveled_codec:ledger_value(),
-                term()) -> term()).
+                dynamic()) -> dynamic()).
```
ThomasArts, line 307: Note that this is OTP26 and later. Not sure that impacts it, but try compile and dialyzer on OTP25

src/leveled_penciller.erl
```diff
@@ -1211,8 +1289,10 @@ handle_cast({complete_shutdown, ShutdownType, From}, State) ->
     {stop, normal, State}.
 
 %% handle the bookie stopping and stop this snapshot
-handle_info({'DOWN', BookieMonRef, process, _BookiePid, _Info},
-	    State=#state{bookie_monref = BookieMonRef}) ->
+handle_info(
+    {'DOWN', BookieMonRef, process, _BookiePid, _Info},
+    State=#state{bookie_monref = BookieMonRef, source_penciller = SrcPCL})
+        when ?IS_DEF(SrcPCL) ->
```
ThomasArts, line 1288: Is there any value of handling the down message with a LOG message of some sort in case we get this when the `source_penciller` is `undefined`. I guess not, because the type system should guarantee that this is not happening?

I think of a situation in which this `gen_server` has been restarted and then the monitor message comes in. ...

src/leveled_penciller.erl
```diff
@@ -1262,7 +1342,9 @@ sst_filename(ManSQN, Level, Count) ->
 %%% Internal functions
 %%%============================================================================
 
--spec update_clerk(pid()|undefined, fun((pid(), term()) -> ok), term()) -> ok.
+-spec update_clerk
+    (pid()|undefined, fun((pid(), atom()) -> ok), atom()) -> ok;
```
ThomasArts: It is a bit more work, but I would try to change this on the caller side.
If we call this function, make sure we call it with a singleton list instead of only an atom?

src/leveled_pmem.erl
```diff
@@ -41,9 +41,12 @@
         cache_full/1
         ]).
 
+% Test functions to ignore for equalizer - due to array issues
+-eqwalizer({nowarn_function, index_performance_test/0}).
+
 -define(MAX_CACHE_LINES, 31). % Must be less than 128
 
--type index_array() :: list(array:array())|[]|none.
+-type index_array() :: list(array:array(binary()))|[]|none.
```
ThomasArts: Why is `[]` a separate item here. That's just an instance of a list of arrays?

ThomasArts:
```suggestion
-type index_array() :: list(array:array(binary()))|none.
```

src/leveled_sst.erl
```diff
@@ -199,27 +206,38 @@
...
-                    :: non_neg_integer()|not_counted,
-            high_modified_date :: non_neg_integer()|undefined,
-            filter_fun :: summary_filter() | undefined,
-            monitor = {no_monitor, 0} :: leveled_monitor:monitor()}).
+    {
+        summary,
+        penciller :: pid() | undefined | false,
+        root_path,
+        filename,
+        read_state :: read_state() | undefined,
```
ThomasArts, line 229: is this really ever going to be `undefined`?
It looks like that whenever you assign `read_state` it is defined and it also is defined initially.
So, one could take away `| undefined` here and all `?IFDEF(RS)` in the code and get the same result??

src/leveled_runner.erl
```diff
@@ -353,11 +353,16 @@ foldobjects_allkeys(SnapFun, Tag, FoldObjectsFun, sqn_order) ->
 
     Folder =
         fun() ->
-            {ok, LedgerSnapshot, JournalSnapshot, AfterFun} = SnapFun(),
-            {ok, JournalSQN} = leveled_inker:ink_getjournalsqn(JournalSnapshot),
+            {ok, LedgerSnapshot, JournalSnapshot, AfterFun} = 
+                case SnapFun() of
+                    {ok, LS, JS, AF} when is_pid(JS) ->
```
ThomasArts, line 358: This is really adding code for the sake of type checking... 
Not sure how to solve it... dependent type for `leveled_bookie:return_snapfun` possibly.

src/leveled_runner.erl
```diff
@@ -630,24 +634,23 @@ accumulate_objects(FoldObjectsFun, InkerClone, Tag, DeferredFetch) ->
                     {B0, K0, _T0} ->
                         {B0, K0}
                 end,
-            JK = {leveled_codec:to_ledgerkey(B, K, Tag), SQN},
+            JK = {leveled_codec:to_objectkey(B, K, Tag), SQN},
             case DeferredFetch of
-                {true, JournalCheck} ->
+                {true, JournalCheck} when MD =/= null ->
```
ThomasArts, line 639: I am not 100% sure this forcing of MD to be not `null` results in the same behaviour when `InJournal` turns out to be false.

src/leveled_codec.erl
```diff
@@ -751,17 +802,28 @@ return_proxy(Tag, ObjMetadata, InkerClone, JournalRef) ->
                         InkerClone,
                         JournalRef}}).
 
+-spec set_status(
+        add|remove, non_neg_integer()|infinity) ->
+            tomb|{active, non_neg_integer()|infinity}.
```
ThomasArts, line 807: I guess, another way of solving this is to add 2 functions and change it at the caller side... 
`get_status_add(TTL) -> {active, TTL} .`  and `get_status_remove() -> tomb`. But that requires a case statement at caller side to extract the IdxObj from the object spec. 
Not sure that that gives a lot, though

ThomasArts (commented):
Looks good, some smaller comments.

include/leveled.hrl
martinsumner: There used to be some unit tests that generated artificial manifest entries, used the owner to identify manifest_entry when asserting. To make this identification easier, the owner was given as a string, rather than spawning PIDs to create artificial manifest entries. 

Going forward, on principle the types should be accurate, even if that takes a lit more effort to generate test cases.

src/leveled_pclerk.erl
martinsumner: The sst_newmerge/10 test function using to include the CountTombs boolean as the final argument - which defaulted to true (this was defaulted in the code, and wasn't configurable by the user).  As part of the refactoring, eqwalizer complained that this could never be `false`.

There were some other options like this, where experimenting with features the code treated things as options that weren't.  To satisfy eqwalizer, I needed to either make it a legitimate option, or remove the code path.  I chose to remove the code path, and so the tests of penciller clerk when `CountTombs = false` became irrelevent.

src/leveled_pmem.erl
martinsumner: Agree - not sure why this exists.  Eqwalizer & dialyzer passes without it

src/leveled_tictac.erl
martinsumner: Indeed

src/leveled_imanifest.erl
martinsumner: There are two manifests - the penciller manifest and the inker manifest.  This is a manifest entry for the inker, where as the one for the penciller is in leveled.hrl.

But, I agree, this is confusing.

src/leveled_sst.erl
martinsumner: Don't think so.  It is at the end of the `starting` state that the read_state is added - it is required to transition to `reader`, but not required within `starting`.

One for further conversation.

src/leveled_runner.erl
martinsumner: The type of object that will be found in accumulate_objects should always be something with metadata - only index entries should not have metadata.  The Tag restricts the object to being a head-only object or an actual object - and in both cases they have metadata.

If there is a deferred fetch the metadata is required.  If there is not a deferred fetch, the metadata is in the value that will be fetched so the metadata is not explicitly required.

src/leveled_runner.erl
martinsumner: As it is an object fold, it needs a snapshot of the journal - other folds don't.  Perhaps there needs to be two types of snap_fun()?  Though perhaps that just might squeeze the problem elsewhere.

src/leveled_penciller.erl
martinsumner: There is a helper function in leveled.hrl, which for OTP < 26 defined dynamic() as any()

src/leveled_penciller.erl
martinsumner: The penciller if it is a primary, will be started as a linked process by the bookie.  If it is a snapshot, it is not necessarily linked to the bookie, but will specifically monitor the bookie on startup.  So the 'DOWN' message from the bookie will only occur in snapshots, and in snapshots the source_penciller must be defined.

src/leveled_penciller.erl
martinsumner: I've expanded the types for clarity

src/leveled_imanifest.erl
martinsumner: Penciller manifest manifest_entry record now moved from global include, and put within leveled_pmanifest module - so that all references to it are now module specific.

src/leveled_sst.erl
martinsumner: Further conversation took place, and there was no obvious neater way of doing this.

src/leveled_codec.erl
martinsumner: Indeed.  I'm happy enough with it as is

PR #455, opened 2024-11-01
Make log functions exportable

martinsumner:
To make it easier to switch to logger in kv_index_tictactree - export the log functions from leveled so that they can be reused


src/leveled_log.erl
```diff
@@ -428,12 +432,15 @@ is_active_level([L|_], L, _) -> true;
...
-    log_timer(LogReference, Subs, StartTime, ?LOG_LEVELS).
+    log_timer(LogReference, Subs, StartTime, ?LOG_LEVELS, ?LOGBASE, backend).
 
-log_timer(LogRef, Subs, StartTime, SupportedLevels) ->
-    {LogLevel, Log} = maps:get(LogRef, ?LOGBASE),
+-spec log_timer(
+    atom(), list(), erlang:timestamp(), list(log_level()), log_base(), atom())
+        -> term().
+log_timer(LogRef, Subs, StartTime, SupportedLevels, LogBase, Tag) ->
+    {LogLevel, Log} = maps:get(LogRef, LogBase, LogBase),
```
ThomasArts: This seems wrong. The default value for the get should either be omitted or be something computed from Supported levels or so. In any case not returning the `?LOGBASE` map and then matching that against a tuple.

src/leveled_log.erl
```diff
@@ -494,8 +501,11 @@ log_test() ->
     log_timer(d0001, [], os:timestamp()).
 
 log_warning_test() ->
-    ok = log(g0001, [], [warning, error]),
-    ok = log_timer(g0001, [], os:timestamp(), [warning, error]).
+    ok = log(g0001, [], [warning, error], ?LOGBASE, backend),
+    ok =
+        log_timer(
+            g0001, [], os:timestamp(), [warning, error], ?LOGBASE, backend
+        ).
```
ThomasArts, line 523: Add a test for the unknown key `wrongkey`:
`??? = log_timer(wrongkey, [], os:timestamp(), [warning, error], ?LOGBASE, backend)`
I think it should return ok and have no side effect action, similar to when the key would be present, but has a log level one is not interested in.

src/leveled_log.erl
```diff
@@ -494,8 +501,11 @@ log_test() ->
...
 
 shouldilog_test() ->
```
ThomasArts, line 539: This tests that the function does not crash. but hardly that you get debug logs?

ThomasArts (commented):
One possibility to crash the log timer... if that is on purpose to find wrong log codes in testing, then fine, but could be more explicit in code

src/leveled_log.erl
martinsumner: I think there should be no default (as with the log function).  If there was a default, it might have the wrong number of substitutions.  So we let it crash if there is a log to a missing log reference.

For leveled, and kv_index_tictactree they both have 100% code coverage in tests (and I maintain this as a requirement), and so there never should be a missing reference get through review.

src/leveled_log.erl
martinsumner: See above.  I've removed the default, and so wrongkey will crash.  I've added a test to assertException to make this explicit.

src/leveled_log.erl
martinsumner: I've removed the debug log test, the shouldilog_test/0 covers this functionality.

ThomasArts (approved):

PR #460, opened 2024-11-26
Mas d34 leveled.i459 partialmerge

### LINKED ISSUE #459: Need for partial merge
When a level is over capacity in the ledger, a file is packed almost at random (in some cases grooming can be an influence) - and that file is fully merged into the level below.

Between level 4 and Level 5 say, there are 384 files at Level 4 and 2304 files at Level 5.  So an average 6 files at level 5 for everyone at level 4 - so on average one would expect to merge into 6 files below.

But what about extreme scenarios.  Imagine a situation with three object buckets `<<"B0">>`, `<<"B1">>` and `<<"B2">>`.  At the start of the database all objects loaded had keys in `<<"B1">>`, but after a point, all updates are a 50/50 mix to buckets <<"B0">> and `<<"B2">>`.

There could be now a scenario where there are 2000 files at Level5 which contain only `<<"B0">>` objects, and amongst the files at Level 4 ones that contain only `<<"B0">>` objects, ones that contain only `<<"B1">>` objects - and one file in between (as each level is ordered) which contains some `<<"B0">>` objects and some `<<"B2">>` objects.

If the file chosen for compaction is the mixed file - it will need to merge into > 2000 files below - although most of those merge events will be taking continuously from the same (lower) side in succession.  This will take a long time - and if updates are ongoing, this will lead to a significant number of `slow_offer` backend pauses.

These events have been rare so far.  Running a perf_SUITE test with 32M objects (much bigger than normal) such an event was triggered though ... and a load which should have taken 10 minutes took 6 hours.  Such severe fluctuations are very much against the principles of what leveled should achieve.  So although this may be a very rare event, it should still be considered a `HIGH` priority bug.

martinsumner:
To resolve https://github.com/martinsumner/leveled/issues/459.

The PR introduces a `max_mergebelow` size which can be a positive integer, or infinity.  It defaults to 24.  This is a startup option, but it isn't expected to be something the end-user is encouraged to tune - the configurability is primarily to aid testing.

As presently, the same algorithm (which is largely random) will pick a file at Level N to merge into Level N + 1, when Level N is overflowing and N is the closest level to the root which is overflowing.  The same process is then used to discover the list of files at Level N + 1 which the chosen merge file covers.  This number can be anywhere from 0 to the file count in Level N + 1. 

If a merge from Level N covers less than `max_mergebelow` files in level N + 1 - the merge will proceed as before.  It is expected that this will occur > 99% of the time in the majority of implementations.

If a merge has >= `max_mergebelow`, the merge will be curtailed when `max_mergebelow div 2` files have been created at that level i.e. the number of additions to be made to Level N + 1 is limited to this value.  In some circumstances, this may merge all the files (e.g. when N + 1 is a basement and the merged file contains a large number of tombstones) - but in most cases there will be a remainder of KV pairs from the merge file at Level N, and some remaining extracted pairs at Level N + 1 as well as some whole files.

The remainder for Level N will then be written as a new file, as well as for Level N + 1 up to the next whole file that has no yet been touched by the merge.

The manifest change will remove the old Level N file, any Level N + 1 files it was merged into, and then add the remainder file at both Level N and Level N + 1.

The backlog that prompted the merge will still exist - as the files in Level N have not been changed.  However, it is likely the next file picked will not be the same one, and will in probability have a lower number of files to merge (as the average is =< 8).  This is why in this case only merging half not all the `max_mergebelow` is required to cease the merge - as it has to be acknowledged that another merge will still be required.

This will stop progress from being halted by long merge jobs, as they will exit out in a safe way after partial completion.  In the case where the majority of files covered  do not require a merge, then those files will be skipped the next time the remainder file is picked up for merge at Level N.

### DIFF
--- a/include/leveled.hrl
+++ b/include/leveled.hrl
@@ -33,6 +33,7 @@
 %%% Non-configurable startup defaults
 %%%============================================================================
 -define(MAX_SSTSLOTS, 256).
+-define(MAX_MERGEBELOW, 24).
 -define(LOADING_PAUSE, 1000).
 -define(LOADING_BATCH, 1000).
 -define(CACHE_SIZE_JITTER, 25).
@@ -109,7 +110,8 @@
                         press_level = ?COMPRESSION_LEVEL :: non_neg_integer(),
                         log_options = leveled_log:get_opts() 
                             :: leveled_log:log_options(),
-                        max_sstslots = ?MAX_SSTSLOTS :: pos_integer(),
+                        max_sstslots = ?MAX_SSTSLOTS :: pos_integer()|infinity,
+                        max_mergebelow = ?MAX_MERGEBELOW :: pos_integer()|infinity,
                         pagecache_level = ?SST_PAGECACHELEVEL_NOLOOKUP
                             :: pos_integer(),
                         monitor = {no_monitor, 0}
--- a/src/leveled_pclerk.erl
+++ b/src/leveled_pclerk.erl
@@ -269,7 +269,6 @@ notify_deletions([Head|Tail], Penciller) ->
 %% to be merged into multiple SSTs at a lower level.  
 %%
 %% SrcLevel is the level of the src sst file, the sink should be srcLevel + 1
-
 perform_merge(Manifest, Src, SinkList, SrcLevel, RootPath, NewSQN, OptsSST) ->
     leveled_log:log(pc010, [leveled_pmanifest:entry_filename(Src), NewSQN]),
     SrcList = [{next, Src, all}],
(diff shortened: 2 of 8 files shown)


martinsumner:
Once an individual vnode or leveled backend reaches 25bn keys in size, then merges into the basement level will increasingly become partial - which may mean multiple merges are required to clear a backlog at Level 6.

At this stage, assuming a ring-size of 512, this amounts to 4.2 trillion keys in the Riak cluster (more if there are a majority of index keys).  For Riak, the answer would be to increase the ring-size.

I assume here that nothing needs to be done about this, as this is much bigger vnode size than anticipated.  At this stage, there are other potential issues running at such untested scale.

src/leveled_pclerk.erl
```diff
@@ -279,72 +278,188 @@ perform_merge(Manifest, Src, SinkList, SrcLevel, RootPath, NewSQN, OptsSST) ->
...
+
+
+-spec split_unexpanded_files(
+    list(merge_maybe_expanded_pointer())) -> 
+        {
+            list(leveled_codec:ledger_kv()|leveled_sst:slot_pointer()),
+            list(leveled_sst:sst_pointer())
+        }. 
+split_unexpanded_files(Pointers) ->
+    split_unexpanded_files(Pointers, [], []).
```
ThomasArts, line 441: Yes, this works. As long as the file pointers `{next, _, _}` are the tail of the list, then this split works, otherwise fails as by design. It does not guarantee that the `ledger_kv` and `slot_pointers` are not occuring in any random order, but that seems fine.

include/leveled.hrl
```diff
@@ -109,7 +110,8 @@
                         press_level = ?COMPRESSION_LEVEL :: non_neg_integer(),
                         log_options = leveled_log:get_opts() 
                             :: leveled_log:log_options(),
-                        max_sstslots = ?MAX_SSTSLOTS :: pos_integer(),
+                        max_sstslots = ?MAX_SSTSLOTS :: pos_integer()|infinity,
+                        max_mergebelow = ?MAX_MERGEBELOW :: pos_integer()|infinity,
```
ThomasArts, line 114: Since you do `div 2` this "user should not touch" value should better not less than 2.
However, since you warn the user not to play with this, if ts fine as it is.

src/leveled_pclerk.erl
```diff
@@ -279,72 +278,188 @@ perform_merge(Manifest, Src, SinkList, SrcLevel, RootPath, NewSQN, OptsSST) ->
...
+        end,
+    {Man1, [Src|SinkManifestRemovals]}.
+
+-spec merge_limit(
+    non_neg_integer(), non_neg_integer(), pos_integer()|infinity)
+        -> pos_integer()|infinity.
+merge_limit(SrcLevel, SinkListLength, MMB) when SrcLevel =< 1; SinkListLength < MMB ->
+    infinity;
+merge_limit(SrcLevel, SinkListLength, MMB) when is_integer(MMB) ->
+    AdditionsLimit = max(1, MMB div 2),
```
ThomasArts, line 332: Ah, user cannot even go below 1. Very good.

src/leveled_pclerk.erl
```diff
@@ -279,72 +278,188 @@ perform_merge(Manifest, Src, SinkList, SrcLevel, RootPath, NewSQN, OptsSST) ->
...
+    Man1 =
+        case L1Additions of
+            [] ->
+                leveled_pmanifest:remove_manifest_entry(
+                    Man0,
+                    NewSQN,
+                    SrcLevel,
+                    Src
+                );
+            PartialFiles ->
```
ThomasArts, line 315: This is always at most one file, isn't it?

ThomasArts (approved):
Elegant

martinsumner:
perfSUITEv9-pr460

The comparison line is an old test of this branch on OTP 24 - and it compares the last 26.2.5 test pre eqwalizer, post eqwalizer and this PR (which includes eqwalizer changes).

The update/load stability and improvements we expect are there.  There is also a mysterious improvement in the GET time.  there is also a mysterious degradation in the query time, which is mainly in smaller volume test runs.

These things I can't explain, unless there is some secondary impact associated with page caches.  i.e. The big merges primarily impacted SST files with index entries, and promoted them in the cache - now the SST files covering object entries are more likely to be in the page cache - so GET/HEAD improve at the cost of QUERY?

src/leveled_pclerk.erl
martinsumner: Yes, though I'm not sure it is important to be explicit about this, as a list of any length would be accepted by replace_manifest_entry/5

Issue #465, opened 2025-01-09
Fold objects by_sqn continues to fold after the Journal snapshot SQN


martinsumner:
Although the fold will not any objects after the Journal snapshot SQN:

https://github.com/martinsumner/leveled/blob/fe9710b47a9e6339b7b4048d9729bf256e49741b/src/leveled_runner.erl#L363-L372

The fold will continue to load batches until the end of the last CDB file in the manifest.  This may cover > 100K objects, and so the end of the fold is delayed.

This caused a Riak issue with handoffs where the remainder after the last batch is only returned to the `riak_core_handoff_sender` after the fold has completed - and if this more than 5 minutes after the last batch was sent, the `riak_core_handoff_receiver` will have timed out, and the handoff does not now complete.  The remainder is not received, and if it is an ownership handoff it has to be started from scratch.

martinsumner:
https://github.com/martinsumner/leveled/pull/466

Issue #462, opened 2024-12-02
Crash output after ct tests


martinsumner:
On completion of the ct tests, sometimes a crash is seen of some leveled_cdb files.

These are leveled_cdb files from the end of the tictac_SUITE basic_headonly test - where they have been compacted, and then the inker is immediately closed:

```
Closing actual store <0.24249.137>

2024-12-02T20:51:28.887 log_level=info log_ref=i0013 db_id=65536 pid=<0.24251.137> File testHO/journal/journal_files/5179_6ca1e6b8-62e0-4395-ac79-0e849e34f1f0 to be removed from manifest

2024-12-02T20:51:28.887 log_level=info log_ref=i0013 db_id=65536 pid=<0.24251.137> File testHO/journal/journal_files/4532_e157acd0-4928-441d-8db8-d92605cc6fb9 to be removed from manifest

2024-12-02T20:51:28.888 log_level=info log_ref=i0016 db_id=65536 pid=<0.24251.137> Writing new version of manifest for manifestSQN=13

2024-12-02T20:51:28.898 log_level=info log_ref=i0005 db_id=65536 pid=<0.24251.137> Inker closing journal for reason close

2024-12-02T20:51:28.898 log_level=info log_ref=i0006 db_id=65536 pid=<0.24251.137> Close triggered with journal_sqn=6250 and manifest_sqn=13

2024-12-02T20:51:28.898 log_level=info log_ref=i0007 db_id=65536 pid=<0.24251.137> Inker manifest when closing is:

2024-12-02T20:51:28.899 log_level=info log_ref=pc005 db_id=65536 pid=<0.24255.137> Penciller's Clerk <0.24255.137> shutdown now complete for reason normal

2024-12-02T20:51:28.899 log_level=info log_ref=p0008 db_id=65536 pid=<0.24254.137> Penciller closing for reason close

2024-12-02T20:51:28.899 log_level=info log_ref=p0010 db_id=65536 pid=<0.24254.137> level zero discarded_count=0 on close of Penciller

2024-12-02T20:51:28.901 log_level=info log_ref=b0003 db_id=65536 pid=<0.24249.137> Bookie closing for reason normal

Trim has reduced journal count from 12 to 5 and 5 after restart
```

The issue here is that close prompts a close of the manifest, however delete_pending files will linger until there next delete_confirmed check.  At this stage they will recognise the inker is not alive and then close:

[...]

PR #467, opened 2025-01-10
Mas d34 leveled.i465 stopfold

martinsumner:
Improve efficiency of fold_objects sqn_order - and hence efficiency of handoffs in Riak

https://github.com/martinsumner/leveled/issues/465

(also includes - https://github.com/martinsumner/leveled/issues/462)

The `riak_SUITE:handoff_close/1` test would previously fail, as there would be around 5s of time processing changes after the fold to the JournalSQN has finished.  It now tests that this is less than 1s (on a fats machine should be < 10ms).

The batch size has also bene reduced to alternate more frequently between pushing disk and pushing CPU.


rebar.config
```diff
@@ -31,8 +31,8 @@
  ]}.
 
 {deps, [
-  {lz4, ".*", {git, "https://github.com/nhs-riak/erlang-lz4", {branch, "nhse-develop-3.4"}}},
-  {zstd, ".*", {git, "https://github.com/nhs-riak/zstd-erlang", {branch, "nhse-develop"}}},
+  {lz4, ".*", {git, "https://github.com/OpenRiak/erlang-lz4", {branch, "openriak-3.4"}}},
+  {zstd, ".*", {git, "https://github.com/OpenRiak/zstd-erlang", {branch, "openriak-3.2"}}},
```
ThomasArts, line 35: just checking... the `3.2` is right here and should not be `3.4`?

ThomasArts (approved):

rebar.config
martinsumner: Yes.  I use openriak-3.2 branch until a change requires the creation of openriak-3.4, and zstd-erlang has not yet required such a change

PR #472, opened 2025-01-20
Add ability to defer fetching a value in the fold - but without requiring Journal check

martinsumner:
…ing check presence

Reduces the cost of using head folds in partition repairs, where the check is not necessary as a fetch failure will be caught and prompt read repair.

### DIFF
--- a/src/leveled_runner.erl
+++ b/src/leveled_runner.erl
@@ -269,11 +269,14 @@ tictactree(SnapFun, {Tag, Bucket, Query}, JournalCheck, TreeSize, Filter) ->
         end,
     {async, Runner}.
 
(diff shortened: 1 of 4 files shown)


src/leveled_runner.erl
stephenwood24-nhs, line 1: Is the spec for `accumulate_objects` correct as DeferredFetch could be `{true, {defer, _}}`
Also is this passed around other functions, if so do their specs need updating?

stephenwood24-nhs (changes requested):

src/leveled_runner.erl
martinsumner: I will update the specs in the leveled_runner module.

stephenwood24-nhs (approved):

PR #475, opened 2025-01-26
SST sub-blocks and block types

martinsumner:
The profile test for most different run-types has `binary_to_term/1` as a significant user of CPU time.

Within `leveled_sst`, each file is made up of slots (normally 256 of them), and each slot is divided into blocks.  The slot normally has five blocks - OuterLeft, InnerLeft, Mid, InnerRight, OuterRight.  There are two types of slots `lookup` and `no_lookup`.  In a `lookup` slot, at least one of the key/values in the slot has a hash as it is expected to be looked up directly.  With `no_lookup` slots, all KV pairs are un-hashed i.e. they are only to be fetched through range queries.

Normally, there are three access requests for information in a slot:

- `get_nth` - i.e. get the KV pair in position N, with N having been found via a hash lookup.
- `get_all` - i.e. get all KVs in the slot - such as in large range folds, bit also to fetch for merge during compaction.
- `get_range` - i.e. get only the KVs within a given range.

The larger the blocks, the more efficient the creation and storage of blocks (due to better compression), and the easier it is to `get_all` (i.e. the more that is fetched within a dingle operation.  The smaller the blocks - the less work to find the nth or a range - as fewer unneeded KV pairs are de-serialised as part of the process.

The aim is to find a way of using smaller blocks, but without paying the penalty at compaction/write time.

The proposed solution is to use multiple block types, with a format tailored for the expected use of the block.  These types of blocks should have minimal changes to enhance efficiency of access.

The block types are:

- Block Type 0 - single block, term_to_binary/1 list of KV pairs, compressed.
- Block Type 1 - a block of 24 KV pairs of type `lookup`, split into four sub-blocks of 6 KV pairs each.
- Block Type 2 - a block of 32 KV pairs of type `lookup`, split into four sub-blocks of 8 KV pairs each.
- Block Type 3 - a block of any length, of type `no_lookup`, which outside of the compressed part has the first and last keys in a mini sub-block appended as a short `term_to_binary/1` object.
- Block Type 4 - as above, but where there are 56 KV pairs in the block, and the block is split into a 24 pair LHS sub-block, an 8 pair mid sub-block and a 24 pair RHS sub-block.

Sub-blocks are all compressed together, they can only be accessed post decompression.  The block includes 2-byte sub-block lengths of every block.  Any block that does not match the requirements of Block Types 1 - 4 (including sub-blocks with length  > 64KB), will use Block Type 0.

For get_nth requests to Block Type 1 & 2, have to decompress the whole block, but only need to deserialise (i.e. `binary_to_term/1`) the relevant block.

For range requests with Block Type 3, the first and last keys can be accessed to determine whether this block and/or neighbouring blocks are required to be deserialised (without having to deserialise the block).  With Block Type 4, the same is true, but when the range is entirely within the block, by deserialising the mid sub-block first it may be possible to avoid deserialising either the left or right sub-blocks.

The Block Types were added in stages, with each stage subjected to perf_SUITE testing in turn, and compared to a control (the current develop-3.4 branch).

- v1 - introduced Block Types 0, 1 and 2;
- v2 - added Block Type 3;
- v3 - added Block Type 4.

perfSUITE-i474

There is a small cost at data load time from the more complex blocks, but this is counter-balanced by significant gains in other operations.

The advantages of block types are largely limited to LZ4 and ZSTD compression, as in native compression the decompression operation is coupled to the deserialisation.

### DIFF
--- a/src/leveled_codec.erl
+++ b/src/leveled_codec.erl
@@ -468,16 +468,16 @@ isvalid_ledgerkey(_LK) ->
 %% false and further results may be required from further ranges.
 endkey_passed(all, _) ->
     false;
-endkey_passed({K1, null, null, null}, {K1, _, _, _}) ->
-    false;
-endkey_passed({K1, K2, null, null}, {K1, K2, _, _}) ->
-    false;
-endkey_passed({K1, K2, K3, null}, {K1, K2, K3, _}) ->
-    false;
-endkey_passed({K1, null}, {K1, _}) ->
+endkey_passed({KQ1, null, null, null}, {KR1, _, _, _}) when KQ1 =/= null ->
+    KQ1 < KR1;
+endkey_passed({K1, KQ2, null, null}, {K1, KR2, _, _}) when KQ2 =/= null ->
+    KQ2 < KR2;
+endkey_passed({K1, K2, KQ3, null}, {K1, K2, KR3, _}) when KQ3 =/= null ->
+    KQ3 < KR3;
+endkey_passed({KQ1, null}, {KR1, _}) when KQ1 =/= null ->
     % See leveled_sst SlotIndex implementation.  Here keys may be slimmed to
     % single binaries or two element tuples before forming the index.
-    false;
+    KQ1 < KR1;
 endkey_passed(null, _) ->
     false;
 endkey_passed(QueryEndKey, RangeEndKey) ->
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -2821,12 +2821,46 @@ next_block_required({_FK, LK}, EndKey) ->
 in_range(KVL, all, all) ->
     KVL;
 in_range(KVL, all, EK) ->
-    lists:takewhile(
-        fun({K, _V}) -> not leveled_codec:endkey_passed(EK, K) end, KVL);
+    before_end(KVL, EK, []);
 in_range(KVL, SK, all) ->
-    lists:dropwhile(fun({K, _V}) -> K < SK end, KVL);
+    after_start(KVL, SK);
 in_range(KVL, SK, EK) ->
-    in_range(in_range(KVL, SK, all), all, EK).
+    before_end(after_start(KVL, SK), EK, []).
+
+before_end(KVL, EK, Acc) when length(KVL) > 12 ->
+    case leveled_codec:endkey_passed(EK, element(1, lists:nth(6, KVL))) of
+        true ->
+            append(
+                Acc,
+                lists:takewhile(
+                    fun({K, _V}) -> not leveled_codec:endkey_passed(EK, K) end,
+                    KVL
+                )
+            );
+        false ->
+            {B, MB} = lists:split(6, KVL),
+            before_end(MB, EK, append(Acc, B))
+    end;
+before_end(KVL, EK, Acc) ->
+    append(
+        Acc,
+        lists:takewhile(
+            fun({K, _V}) -> not leveled_codec:endkey_passed(EK, K) end,
+            KVL
+        )
+    ).
+
+after_start(KVL, SK) when length(KVL) > 24 ->
+    case element(1, lists:nth(12, KVL)) < SK of
+        true ->
+            {_B, MB} = lists:split(12, KVL),
+            after_start(MB, SK);
+        false ->
+            lists:dropwhile(fun({K, _V}) -> K < SK end, KVL)
+    end;
+after_start(KVL, SK) ->
+    lists:dropwhile(fun({K, _V}) -> K < SK end, KVL).
+
 
 crc_check_slot(FullBin) ->
     <<CRC32PBL:32/integer,
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -4637,7 +4637,7 @@ corrupted_block_rangetester(BlockMethod, TestCount) ->
             no_lookup, BlockMethod, lists:sublist(KVL1, 41, 20)),
     B4 =
         leveled_sstblock:serialise_block(
-            np_lookup, BlockMethod, lists:sublist(KVL1, 61, 20)),
+            no_lookup, BlockMethod, lists:sublist(KVL1, 61, 20)),
     B5 =
         leveled_sstblock:serialise_block(
             no_lookup, BlockMethod, lists:sublist(KVL1, 81, 20)),
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -134,7 +134,7 @@
 
 -export([in_range/3]).
 
--export([hmac/1, append/2, append/3, append/4, filterby_midblock/2]).
+-export([hmac/1, filterby_midblock/2]).
 
 -record(slot_index_value,
         {slot_id :: integer(),
--- a/src/leveled_sst.erl
+++ b/src/leveled_sst.erl
@@ -2689,90 +2689,103 @@ blocks_required(
         leveled_sstblock:get_topandtail(MidBlock, BlockMethod),
     case filterby_midblock({Top, Tail}, {StartKey, EndKey}) of
         empty ->
-            append(
-                in_range(
+            in_range(
+                append(
                     leveled_sstblock:get_all(B1, BlockMethod),
-                    StartKey,
-                    EndKey
-                ),
-                in_range(
                     leveled_sstblock:get_all(B2, BlockMethod),
-                    StartKey,
-                    EndKey
-                ),
-                in_range(
                     leveled_sstblock:get_all(B4, BlockMethod),
-                    StartKey,
-                    EndKey
+                    leveled_sstblock:get_all(B5, BlockMethod)
                 ),
-                in_range(
-                    leveled_sstblock:get_all(B5, BlockMethod),
-                    StartKey,
-                    EndKey
-                )
+                StartKey,
+                EndKey
             );
         all_blocks ->
             append(
-                get_lefthand_blocks(B1, B2, BlockMethod, StartKey),
+                in_range(
+                    get_lefthand_blocks(
+                        B1, B2, BlockMethod, StartKey, EndKey),
+                    StartKey,
+                    all
+                ),
                 MidBlockFetchFun(),
-                get_righthand_blocks(B4, B5, BlockMethod, EndKey)
+                in_range(
+                    get_righthand_blocks(
+                        B4, B5, BlockMethod, StartKey, EndKey),
+                    all,
+                    EndKey
+                )
             );
         lt_mid ->
             in_range(
-                get_lefthand_blocks(B1, B2, BlockMethod, StartKey),
-                all,
+                get_lefthand_blocks(
+                    B1, B2, BlockMethod, StartKey, EndKey),
+                StartKey,
                 EndKey);
         le_mid ->
-            append(
-                get_lefthand_blocks(B1, B2, BlockMethod, StartKey),
-                in_range(MidBlockFetchFun(), all, EndKey)
+            in_range(
+                append(
+                    get_lefthand_blocks(
+                        B1, B2, BlockMethod, StartKey, EndKey),
+                    MidBlockFetchFun()
+                ),
+                StartKey,
+                EndKey
             );
         mid_only ->
             in_range(MidBlockFetchFun(), StartKey, EndKey);
         ge_mid ->
-            append(
-                in_range(MidBlockFetchFun(), StartKey, all),
-                get_righthand_blocks(B4, B5, BlockMethod, EndKey)
+            in_range(
+                append(
+                    MidBlockFetchFun(),
+                    get_righthand_blocks(
+                        B4, B5, BlockMethod, all, EndKey)
+                ),
+                StartKey,
+                EndKey
             );
         gt_mid ->
             in_range(
-                get_righthand_blocks(B4, B5, BlockMethod, EndKey),
+                get_righthand_blocks(
+                    B4, B5, BlockMethod, StartKey, EndKey),
                 StartKey,
-                all)
+                EndKey
+            )
     end.
 
-get_lefthand_blocks(B1, B2, BlockMethod, StartKey) ->
+get_lefthand_blocks(B1, B2, BlockMethod, StartKey, EndKey) ->
     {Top, Tail, InnerLeftBlockFetchFun} =
         leveled_sstblock:get_topandtail(B2, BlockMethod),
     case previous_block_required({Top, Tail}, StartKey) of
         true ->
             append(
-                in_range(
-                    leveled_sstblock:get_all(B1, BlockMethod),
-                    StartKey,
-                    all
-                ),
-                InnerLeftBlockFetchFun()
+                leveled_sstblock:get_all(B1, BlockMethod),
+                case this_leftblock_required({Top, Tail}, EndKey) of
+                    true ->
+                        InnerLeftBlockFetchFun();
+                    _ ->
+                        []
+                end
             );
         false ->
-            in_range(InnerLeftBlockFetchFun(), StartKey, all)
+            InnerLeftBlockFetchFun()
     end.
 
-get_righthand_blocks(B4, B5, BlockMethod, EndKey) ->
+get_righthand_blocks(B4, B5, BlockMethod, StartKey, EndKey) ->
     {Top, Tail, InnerRightBlockFetchFun} =
         leveled_sstblock:get_topandtail(B4, BlockMethod),
     case next_block_required({Top, Tail}, EndKey) of
         true ->
             append(
-                InnerRightBlockFetchFun(),
-                in_range(
-                    leveled_sstblock:get_all(B5, BlockMethod),
-                    all,
-                    EndKey
-                )
+                case this_rightblock_required({Top, Tail}, StartKey) of
+                    true ->
+                        InnerRightBlockFetchFun();
+                    _ ->
+                        []
+                end,
+                leveled_sstblock:get_all(B5, BlockMethod)
             );
         false ->
-            in_range(InnerRightBlockFetchFun(), all, EndKey)
+            InnerRightBlockFetchFun()
     end.
 
 filterby_midblock({not_present, not_present}, _RangeKeys) ->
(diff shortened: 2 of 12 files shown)


martinsumner:
Look at some before/after on some profiles:

For HEAD requests - in `develop-3.4`:

```bash
FUNCTION                                  CALLS        %      TIME  [uS / CALLS]
leveled_bookie:fetch_head/4                600000     0.53    205622  [      0.34]
zstd:quick_compress/2                        8285     0.55    212680  [     25.67]
leveled_ebloom:match_hash/3               1797546     0.59    226628  [      0.13]
os:timestamp/0                            3500341     0.63    243308  [      0.07]
leveled_sst:spawn_check_block/3            600942     0.63    243536  [      0.41]
array:get_1/3                             5687061     0.64    245694  [      0.04]
prim_file:read_nif/2                        10389     0.64    246573  [     23.73]
rand:seed_put/1                           3041506     0.64    247047  [      0.08]
leveled_pmanifest:key_lookup_level/3      3693943     0.68    263843  [      0.07]
dict:find_val/2                           6740486     0.70    269568  [      0.04]
leveled_penciller:fetch/5                 2500535     0.74    284133  [      0.11]
ets:lookup/2                               600000     0.74    287074  [      0.48]
gen_statem:loop_receive/3                  641715     0.78    301739  [      0.47]
gen_server:loop/7                         1206105     0.85    327012  [      0.27]
leveled_bookie:handle_call/3               600000     0.89    341644  [      0.57]
erlang:demonitor/2                        1847606     0.89    345203  [      0.19]
crypto:hash_nif/2                          600000     0.93    356955  [      0.59]
erlang:spawn_link/3                        600942     1.04    400380  [      0.67]
leveled_sst:fetch/12                       635447     1.17    451786  [      0.71]
leveled_tree:search/3                     1935982     1.27    489096  [      0.25]
rand:uniform/1                            3041470     1.28    491901  [      0.16]
leveled_tree:iterator_from/3              6901621     1.34    515073  [      0.07]
gen:reply/2                               1847497     1.36    524457  [      0.28]
leveled_tree:lookup_best/2               12581788     2.39    921727  [      0.07]
gen:do_call/4                             1847598     2.65   1020798  [      0.55]
prim_file:pread_nif/3                      601071     4.83   1862560  [      3.10]
leveled_sst:find_posint/4                81953690     7.89   3042318  [      0.04]
erlang:binary_to_term/1                    609836    18.12   6988108  [     11.46]
zstd:quick_decompress/1                    608134    18.42   7105779  [     11.68]
--------------------------------------  ---------  -------  --------  [----------]
Total:                                  292137232  100.00%  38572591  [      0.13]
```

This is a measure across 600K HEAD requests.  The binary_to_term/1 calls are slightly greater than 600K (due to bloom misses at higher levels).  At 11.46 microseconds per call, this represents 18.12% of CPU - this indicates about 64 microseconds pre request overall.  This is already fairly fast!  In a full-load test, there's probably about 50K HEAD requests per second (on a 64 CPU core cluster) - this amounts to 3.2 CPU.  However, Graviton CPU can be 2-3 times slower than Apple ARM - so this might equate to about 6 - 9 CPU overall.

In This PR (v3):

```bash
FUNCTION                                  CALLS        %      TIME  [uS / CALLS]
--------                                  -----  -------      ----  [----------]
leveled_ebloom:check_hash/2             1598588     0.57    200518  [      0.13]
leveled_bookie:fetch_head/4              600000     0.59    208694  [      0.35]
leveled_sst:spawn_check_block/3          600472     0.65    229364  [      0.38]
array:get_1/3                           5468842     0.69    240788  [      0.04]
erlang:term_to_binary/1                   94235     0.70    245557  [      2.61]
rand:seed_put/1                         3056477     0.70    247201  [      0.08]
os:timestamp/0                          3545403     0.73    255183  [      0.07]
dict:find_val/2                         6533961     0.74    259456  [      0.04]
ets:lookup/2                             600000     0.76    268106  [      0.45]
leveled_pmanifest:key_lookup_level/3    4049154     0.77    271327  [      0.07]
leveled_penciller:fetch/5               2255575     0.78    273498  [      0.12]
leveled_ebloom:match_hash/3             2413756     0.82    288348  [      0.12]
gen_statem:loop_receive/3                656875     0.85    299237  [      0.46]
crypto:hash_nif/2                        600000     0.92    324455  [      0.54]
gen_server:loop/7                       1206200     0.93    325891  [      0.27]
prim_file:read_nif/2                      10627     0.96    338517  [     31.85]
leveled_bookie:handle_call/3             600000     0.97    341690  [      0.57]
erlang:spawn_link/3                      600472     1.08    377623  [      0.63]
erlang:demonitor/2                      1863235     1.21    424934  [      0.23]
leveled_tree:search/3                   1705867     1.24    434804  [      0.25]
leveled_sst:fetch/12                     650292     1.34    470941  [      0.72]
leveled_tree:iterator_from/3            6235057     1.38    485739  [      0.08]
rand:uniform/1                          3056442     1.43    501491  [      0.16]
gen:reply/2                             1862800     1.44    506220  [      0.27]
zstd:quick_compress/2                     24645     1.65    580774  [     23.57]
leveled_tree:lookup_best/2             10978812     2.44    856074  [      0.08]
gen:do_call/4                           1863221     2.94   1030688  [      0.55]
prim_file:pread_nif/3                    600795     4.26   1496235  [      2.49]
erlang:binary_to_term/1                  682331     6.65   2332533  [      3.42]
leveled_sst:find_posint/4              80416603     8.68   3045895  [      0.04]
zstd:quick_decompress/1                  623940    20.50   7193860  [     11.53]
------------------------------------  ---------  -------  --------  [----------]
Total:                                296101490  100.00%  35100466  [      0.12]
```

The speed improvement is almost entirely in `binary_to_term/1`.  This amounts to about 8 microseconds per request - so although the relative change is significant, the absolute is small.  This may reduce the CPU load by 1 Graviton CPU core in a 64 CPU core cluster - so the overall impact on throughput may not be huge.

martinsumner:
For mini-query (which make up the majority of our Riak 2i calls) - the profile in `develop-3.4`:

```
FUNCTION                                               CALLS        %      TIME  [uS / CALLS]
--------                                               -----  -------      ----  [----------]
gen_server:loop/7                                     601270     0.77    107676  [      0.18]
erlang:put/2                                         1002099     0.83    114951  [      0.11]
lists:reverse/1                                      2555116     0.85    118200  [      0.05]
leveled_tree:idxtlookup_range_end/5                   491122     0.85    118435  [      0.24]
leveled_sst:'-in_range/3-fun-0-'/2                   1713214     0.95    132687  [      0.08]
leveled_tree:iterator_from/3                         1658458     1.00    139751  [      0.08]
lists:map/2                                          1675746     1.02    141455  [      0.08]
gen_statem:loop_receive/3                             380725     1.06    148041  [      0.39]
leveled_penciller:find_nextkeys/6                    1971483     1.18    164756  [      0.08]
lists:takewhile_1/2                                  1713658     1.31    182793  [      0.11]
leveled_tree:'-idxtlookup_range_start/4-fun-0-'/2    2185653     1.39    192895  [      0.09]
erlang:demonitor/2                                    981965     1.74    241946  [      0.25]
lists:splitwith_1/3                                  3169859     2.14    297709  [      0.09]
prim_file:pread_nif/3                                 190355     2.53    352076  [      1.85]
gen:reply/2                                           881755     2.74    380835  [      0.43]
lists:last/2                                        11954019     3.18    442291  [      0.04]
leveled_codec:endkey_passed/2                        3911145     3.41    474724  [      0.12]
lists:dropwhile_1/2                                  6891345     3.68    512590  [      0.07]
gen:do_call/4                                         881756     3.76    523703  [      0.59]
leveled_sst:'-in_range/3-fun-1-'/2                   6789498     4.45    619299  [      0.09]
zstd:quick_decompress/1                               334990     5.87    817344  [      2.44]
erlang:binary_to_term/1                               334990    22.06   3070470  [      9.17]
-------------------------------------------------  ---------  -------  --------  [----------]
Total:                                             118749678  100.00%  13917031  [      0.12]
```

With a ring-size of 512, there will be 172 mini-queries per 2i query.  This is a measure of 60K queries - so that is 52 microseconds per query on binary_to_term.  But that would translate to around 9ms per query when multiplied across a whole 2i query.  Typically we are querying at 60-100 queries per second during a load test: so (applying a Graviton to Apple multiple) this could equate to 2 CPU cores ob binary_to_term/1 alone.

In this PR (v3):

```
FUNCTION                                               CALLS        %      TIME  [uS / CALLS]
--------                                               -----  -------      ----  [----------]
gen_server:loop/7                                     600036     0.84    112349  [      0.19]
leveled_sst:filterby_midblock/2                       550333     0.86    115618  [      0.21]
leveled_tree:idxtlookup_range_start/4                 580031     0.89    119693  [      0.21]
leveled_sst:expand_list_by_pointer/5                  560044     0.94    126685  [      0.23]
erlang:crc32/1                                        976055     1.03    138433  [      0.14]
lists:reverse/1                                      3000151     1.11    149533  [      0.05]
leveled_tree:idxtlookup_range_end/5                   580179     1.17    156882  [      0.27]
lists:map/2                                          1942455     1.44    193448  [      0.10]
leveled_tree:iterator_from/3                         2105299     1.54    207430  [      0.10]
leveled_penciller:find_nextkeys/6                    2240069     1.55    208394  [      0.09]
gen_statem:loop_receive/3                             560059     1.68    226027  [      0.40]
erlang:demonitor/2                                   1160063     2.07    278425  [      0.24]
leveled_codec:endkey_passed/2                        3785157     2.42    325500  [      0.09]
leveled_tree:'-idxtlookup_range_start/4-fun-0-'/2    3700133     2.52    338544  [      0.09]
lists:splitwith_1/3                                  4772390     3.56    478270  [      0.10]
gen:reply/2                                          1060059     3.62    485802  [      0.46]
prim_file:pread_nif/3                                 280022     3.97    533626  [      1.91]
gen:do_call/4                                        1060060     5.00    671856  [      0.63]
zstd:quick_decompress/1                               291800     6.04    811577  [      2.78]
erlang:binary_to_term/1                              1086036    10.60   1423099  [      1.31]
-------------------------------------------------  ---------  -------  --------  [----------]
Total:                                             120754070  100.00%  13428992  [      0.11]
```

There are 3 x as many binary_to-term/1 calls, but they are 80% cheaper.  However, there are also a number of other functions where savings have been made - the functions to find and trim a range in particular.

In this run, there were 50% more calls to prim_file:pread_nif/3 than in `develop-3.4` code.  This would indicate some kind-of trade-off to disk requests ... however, I cannot reason how any additional reads are taking place. The changes are to the subdivision of a block within a slot - pread is used to read the whole slot in both versions of the code.

There is some randomness in the mini-query test run - it may just be that there was more work done in the second test run.  This would correlate with the call count on other functions such as `leveled_penciller:find_nextkeys/6 `.

A second run of the profile confirms that variance:

```
FUNCTION                                               CALLS        %      TIME  [uS / CALLS]
--------                                               -----  -------      ----  [----------]
leveled_tree:idxtlookup_range_start/4                 494042     0.71    100804  [      0.20]
erlang:put/2                                         1000560     0.79    113588  [      0.11]
gen_server:loop/7                                     600348     0.82    117019  [      0.19]
lists:foldl_1/3                                      1800990     0.83    118122  [      0.07]
leveled_tree:idxtlookup_range_end/5                   464077     0.83    118174  [      0.25]
lists:takewhile_1/2                                  1383375     0.90    128403  [      0.09]
lists:map/2                                          1684615     1.04    148521  [      0.09]
lists:nth_1/2                                        3809152     1.05    149659  [      0.04]
gen_statem:loop_receive/3                             387768     1.09    155409  [      0.40]
leveled_tree:iterator_from/3                         1793539     1.17    166859  [      0.09]
leveled_penciller:find_nextkeys/6                    1981752     1.26    179533  [      0.09]
lists:reverse/1                                      4110935     1.32    188686  [      0.05]
erlang:demonitor/2                                    988085     1.77    252762  [      0.26]
leveled_tree:skpllookup_to_range/4                    900495     1.87    268044  [      0.30]
lists:dropwhile_1/2                                  3620143     1.93    276110  [      0.08]
leveled_tree:'-idxtlookup_range_start/4-fun-0-'/2    2816118     1.95    278437  [      0.10]
leveled_tree:'-skpllookup_to_range/4-fun-0-'/2       2963536     2.05    292935  [      0.10]
lists:splitwith_1/3                                  3772645     2.41    344869  [      0.09]
gen:reply/2                                           888029     2.78    397785  [      0.45]
prim_file:pread_nif/3                                 193877     2.87    409872  [      2.11]
leveled_codec:endkey_passed/2                        5494432     3.71    530281  [      0.10]
gen:do_call/4                                         888030     3.99    570870  [      0.64]
zstd:quick_decompress/1                               205567     4.03    575605  [      2.80]
leveled_tree:skplfold_range/4                        9011310     5.72    818457  [      0.09]
erlang:binary_to_term/1                               824312    11.77   1683317  [      2.04]
-------------------------------------------------  ---------  -------  --------  [----------]
Total:                                             131152540  100.00%  14296862  [      0.11]
```

src/leveled_codec.erl
```diff
@@ -468,16 +468,16 @@ isvalid_ledgerkey(_LK) ->
...
 endkey_passed(all, _) ->
     false;
-endkey_passed({K1, null, null, null}, {K1, _, _, _}) ->
-    false;
-endkey_passed({K1, K2, null, null}, {K1, K2, _, _}) ->
-    false;
-endkey_passed({K1, K2, K3, null}, {K1, K2, K3, _}) ->
-    false;
-endkey_passed({K1, null}, {K1, _}) ->
+endkey_passed({KQ1, null, null, null}, {KR1, _, _, _}) when KQ1 =/= null ->
```
ThomasArts, line 471: be aware that:
```
null < <<"a">>.
true
```
Thus `endkey_passed({nul, null, null, null}, {null, null, null, <<"a">>})`  drops through to the last clause and results in `true` whereas in the old version it would return a `false`.

martinsumner: null means don't care in this case, so if we don't care about all elements of the tuple, then we should use `all`, which will return a false because of the first clause.

src/leveled_sst.erl
```diff
@@ -2669,69 +2682,115 @@ binaryslot_trimmed(
...
+    {Top, Tail, InnerLeftBlockFetchFun} =
+        leveled_sstblock:get_topandtail(B2, BlockMethod),
+    case previous_block_required({Top, Tail}, StartKey) of
         true ->
-            in_range(deserialise_block(B1, PressMethod), StartKey, all)
-            ++ BlockList2;
+            case this_leftblock_required({Top, Tail}, EndKey) of
+                true ->
+                    append(
+                        leveled_sstblock:get_all(B1, BlockMethod),
```
ThomasArts, line 2764: The name `get_all` in that module does no longer refer to "deserialize", so one wonders whether the result is deserialized. When looking in the code for leveled_sstblock, it clearly is for version 0 blocks, but harder to distinguish for version 1 blocks (see comment in file).
I am not sure this fits the bill, but could one rename `get_all` to something making clear that what type of objects one gets returned (leveled_codec:ledger_kv() according to spec).
`get_all_deserialized`?

src/leveled_sst.erl
```diff
@@ -2669,69 +2682,115 @@ binaryslot_trimmed(
...
         true ->
-            BlockList4
-            ++ in_range(deserialise_block(B5, PressMethod), all, EndKey);
+            case this_rightblock_required({Top, Tail}, StartKey) of
+                true ->
+                    append(
+                        InnerRightBlockFetchFun(all),
+                        leveled_sstblock:get_all(B5, BlockMethod)
+                    );
+                _ ->
```
ThomasArts:
```suggestion
                false ->
```

src/leveled_sstblock.erl
```diff
@@ -0,0 +1,799 @@
+%% -------- SST Block Functions ---------
+%%
+%% Functions to serialise and deserialise blocks in leveled_sst
```
ThomasArts: This comment suggests an API with serialize and deserialize functions.
`serialize_block` says it all, but it is not clear that the other functions deserialize or possibly partly deserialize.
Possibly it would be nice to reflect that in their name, such as `get_all_deserialized`.
In particular when it comes to version 0 blocks, that is exactly what get_all would do.

From a property point of view, it is right to not call the functions `deserialize`, because there is no roundtrip from serialization to deserialization and back.

src/leveled_sstblock.erl
```diff
@@ -0,0 +1,799 @@
...
+).
+
+%%%============================================================================
+%%% API
+%%%============================================================================
+
+
+-spec serialise_block(
+    lookup|no_lookup,
+    {0|1, leveled_sst:press_method()},
```
ThomasArts:
```suggestion
    {leveled_sst:block_version(), leveled_sst:press_method()},
```
To make it more consistent and to avoid future changes in just one place

src/leveled_sstblock.erl
```diff
@@ -0,0 +1,799 @@
...
+    CheckedSize = byte_size(CheckedBlock),
+    <<TypedBlock:(CheckedSize - 1)/binary, Type:8/integer>> = CheckedBlock,
+    get_nth_item(Type, N, TypedBlock, PressMethod).
+
+-spec get_nth_item(
+    block_type(), pos_integer(), binary(), leveled_sst:press_method()) ->
+        leveled_codec:ledger_kv().
+get_nth_item(Type, N, TypedBlock, PM)
+        when Type == ?BLOCK_TYPE0, (PM == zstd orelse PM == lz4) ->
+    lists:nth(N, binary_to_term(decompress_block(TypedBlock, PM)));
```
ThomasArts: This is the same as 
```
    lists:nth(N, deserialise_checkedblock(TypedBlock, PM));
```
But I guess you want to consequently use `term_to_binary` to show what is going on?
In some cases you don't want to do term_to_binary on the complete term and this makes that more explicit.

ThomasArts (approved):
Looks good. It's quite a lot of extra code to get some optimization, but it paid of.

martinsumner:
So running a full volume test with Riak - this change is enough to make a 5.4% uplift in throughput:

Throughput Comparison - 8KB BV3

src/leveled_sstblock.erl
martinsumner: Updated

src/leveled_sstblock.erl
martinsumner: Change made

src/leveled_sst.erl
martinsumner: I've tried to adjust the comments to the top to make this a bit easier to understand

PR #479, opened 2025-08-07
Standardise formatting using erlfmt

martinsumner:
Adding check of format to github actions

### DIFF
--- a/src/leveled_runner.erl
+++ b/src/leveled_runner.erl
@@ -1,67 +1,68 @@
 %% -------- RUNNER ---------
 %%
-%% A bookie's runner would traditionally allow remote actors to place bets 
-%% via the runner.  In this case the runner will allow a remote actor to 
+%% A bookie's runner would traditionally allow remote actors to place bets
+%% via the runner.  In this case the runner will allow a remote actor to
 %% have query access to the ledger or journal.  Runners provide a snapshot of
-%% the book for querying the backend.  
+%% the book for querying the backend.
 %%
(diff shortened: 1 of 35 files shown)


src/leveled_runner.erl
```diff
@@ -420,7 +420,12 @@ foldobjects_allkeys(SnapFun, Tag, FoldObjectsFun, sqn_order) ->
                         ),
                     % Need to check that we have not folded past the point
                     % at which the snapshot was taken
-                    (JournalSQN >= SQN) and (CheckSQN == current)
+                    case {JournalSQN, CheckSQN} of
```
ThomasArts: No problem, but I dislike the match with the bound variable JournalSQN. Similar for SnapSQN above, but here scope is larger and harder to see bindings.
An '_' in the match would possibly be fine, but why not:
```
   case CheckSQN of
       current -> JournalSQN >= SQN;
       _ -> false
   end
```
Where that is basically an `andalso`, but I understand that that did not work for the formatter.

ThomasArts (approved):

src/leveled_runner.erl
martinsumner: This was another elp/vs-code issue with the syntax highlighting, rather than a formatter issue.  I'm happy with your suggestion.

martinsumner: checked again, and the issue was with the (unnecessary) parenthesis in elp:

```
JournalSQN >= SQN andalso CheckSQN == current
```

is OK

PR #483, opened 2025-11-08
Switch to using domain filtering not log_type

martinsumner:
This looks like it is a more correct way of doing this.

Given we have no agreed hierarchy as part of OpenRiak, will use [backend, leveled] as a start.

### DIFF
--- a/include/leveled.hrl
+++ b/include/leveled.hrl
@@ -88,18 +88,18 @@
 
 -define(IS_DEF(Attribute), Attribute =/= undefined).
 
--define(LOG_LOCATION,
-    #{
-        mfa => {?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY},
-        line => ?LINE,
-        file => ?FILE}
-    ).
+-define(LOG_LOCATION, #{
+    mfa => {?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY},
+    line => ?LINE,
+    file => ?FILE
+}).
 
 -define(STD_LOG(LogRef, Subs),
     erlang:apply(
         logger,
         macro_log,
-        [?LOG_LOCATION|leveled_log:log(LogRef, Subs)])
+        [?LOG_LOCATION | leveled_log:log(LogRef, Subs)]
+    )
 ).
 
 -define(RND_LOG(LogRef, Subs, StartTime, RandomProb),
@@ -108,7 +108,7 @@
             erlang:apply(
                 logger,
                 macro_log,
-                [?LOG_LOCATION|leveled_log:log_timer(LogRef, Subs, StartTime)]
+                [?LOG_LOCATION | leveled_log:log_timer(LogRef, Subs, StartTime)]
             );
         false ->
             ok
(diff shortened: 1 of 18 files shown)


include/leveled.hrl
```diff
@@ -81,11 +83,95 @@
...
+
+-define(STD_LOG(LogRef, Subs),
+    ?STD_LOG_INT(
+        element(1, leveled_log:get_log(LogRef)),
+        LogRef,
+        Subs,
+        leveled_log:get_opts()
+    )
+).
+
```
ThomasArts, line 105:
```suggestion

%% Erlang apply is used because  a variable list of arguments is provided
```

src/leveled_log.erl
```diff
@@ -405,6 +404,10 @@ set_databaseid(DBid) when is_integer(DBid) ->
     UpdLO = LO#log_options{database_id = DBid},
     save(UpdLO).
 
+-spec get_log(atom()) -> {log_level(), binary()}.
```
ThomasArts: It would be nice with a function `get_log_level(LogRef)` in order to not have to write `elements(1, get_log(LogRef)`

include/leveled.hrl
```diff
@@ -81,11 +83,95 @@
...
 %%%============================================================================
 
 -define(IS_DEF(Attribute), Attribute =/= undefined).
 
+-define(LOG_LOCATION, #{
+    mfa => {?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY},
+    line => ?LINE,
+    file => ?FILE
+}).
+
```
ThomasArts, line 96: Is `leveled.hrl` only meant to be included in the source code of this project? That is, are there users that would do
`-include_lib("leveled/include/leveled.hrl")` ??
If users are expected to use this include, then log messages probably should be in separate header file and included in all srcs here.
If this is only for internal use, then one could add ?DOMAIN as macro `[backend, leveled]` and use that in places there now hardcoded `[backend, leveled]` shows up.

ThomasArts (approved):

include/leveled.hrl
martinsumner: No.  There is some overlap in use with https://github.com/martinsumner/kv_index_tictactree/pull/133 - but that does not involve including this file.

In kv_index_tictactree we pass a different domain in, but into leveled_log:log.

Do you think it would be better to extend the ?LOCATION map so that it includes `domain => ?DOMAIN`, so that it doesn't need to be added separately in leveled_log?

ThomasArts: Location is something that is defined by the logger API for normal logger calls:
https://www.erlang.org/doc/apps/kernel/logger.html#miscellaneous-api-functions
I would advise not to add something to it in order to stay as close as possible to that API. This is importnat if someone decides to make their own formatter.
Domain is also something semantically valid that people may use in their custom filters. So better to keep them separate

ThomasArts: Revisiting the maco's you define... you could have an optional 3rd argument STD_LOG macro that takes Metadata `#{domain => ...}`  as third argument. This can then pass the domain on to the internal log function... 
That would be in line with what programmers that do know logger well would expect.

src/leveled_log.erl
martinsumner: Done

include/leveled.hrl
martinsumner: For now simply added the ?DOMAIN macro.  Will extend the log macro should a more flexible domain setting be required in the future.

PR #484, opened 2025-12-16
Don't require regex to be compiled (spec change only)

martinsumner:
And don't compile regex in evaluation expressions


martinsumner:
https://github.com/OpenRiak/riak_kv/issues/102

ThomasArts (approved):

PR #486, opened 2026-01-20
Rename any cdb files outside of the manifest

martinsumner:
Rename them as *.bak files.  This aligns with the ledger, which does the same thing at startup.

These files are expected when a stop or crash happens during a journal compaction.  In this case, files may be created - but not added to the manifest.  These orphaned files will be in the post_compact folder.

It is less expected for folders in the non-compact area, but can still happen when there is a crash during the delete_pending state.

The garbage is not collected (deleted) - but this just makes it easier for OS admins to clear it, as they don't need erlang functions, *.bak files can be assumed to be garbage.

### DIFF
--- a/src/leveled_inker.erl
+++ b/src/leveled_inker.erl
@@ -137,6 +137,7 @@
 -define(WASTE_FP, "waste").
 -define(JOURNAL_FILEX, "cdb").
 -define(PENDING_FILEX, "pnd").
+-define(ARCHIVE_FILEX, "bak").
 -define(TEST_KC, {[], infinity}).
 -define(SHUTDOWN_LOOPS, 10).
 -define(SHUTDOWN_PAUSE, 10000).
@@ -682,7 +683,7 @@ handle_call(roll, _From, State = #state{is_snapshot = Snap}) when
             }}
     end;
 handle_call(
-    {backup, BackupPath}, _from, State
+    {backup, BackupPath}, _From, State
 ) when
     State#state.is_snapshot == true
 ->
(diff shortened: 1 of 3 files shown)


src/leveled_inker.erl
```diff
@@ -699,20 +700,21 @@ handle_call(
                     ExtendedBaseFN = BaseFN ++ "." ++ ?JOURNAL_FILEX,
                     BackupName = filename:join(BackupJFP, BaseFN),
                     true = leveled_cdb:finished_rolling(PidR),
-                    case
+                    Link =
                         file:make_link(
                             FN ++ "." ++ ?JOURNAL_FILEX,
                             BackupName ++ "." ++ ?JOURNAL_FILEX
```
ThomasArts, line 708: Probably because this is a link, but given the newly added archive file extension, I would expect here:
```
                            BackupName ++ "." ++ ?ARCHIVE_FILEX
```

src/leveled_inker.erl
```diff
@@ -1273,51 +1275,93 @@ open_all_manifest([], RootPath, CDBOpts) ->
...
-                OpenedTail, {HeadSQN, HeadFN, HeadW, HeadLK}, true
-            )
-    end.
+                    OpenedTail, {HeadSQN, HeadFN, HeadW, HeadLK}, true
+                )
+        end,
+    lists:foreach(
+        fun(FN) ->
+            NewName =
+                lists:sublist(FN, length(FN) - 4) ++ "." ++ ?ARCHIVE_FILEX,
```
ThomasArts: What is happening here? I assume a kind of deleting the 3 character extension and replace it?
If so, then using the filename functions is more explanatory and less error prone:

```suggestion
                filename:flatten([filename:rootname(FN), "." ++ ?ARCHIVE_FILEX]),
```
Using iolists helps to not get in the situation the FN is a binary and '++' fails. But if you assume charlists later on, then deep list may fail.

src/leveled_inker.erl
ThomasArts: a little comment in the code? Backup is something else than archive?

ThomasArts (commented):
Please double check file renaming, I didn't get what actually happened there.

src/leveled_inker.erl
martinsumner: Comment added to the definition of the file extension

ThomasArts (approved):

PR #487, opened 2026-02-12
leveled_bookie:status/1 (#17)

martinsumner:
Return a map of status information about the bookie (when the bookie is not a snapshot).

The content of the status may change in future releases.

---------


martinsumner:
Reviewed - https://github.com/OpenRiak/leveled/pull/17

PR #488, opened 2026-02-12
Mas d34 masl.i480 asyncput

martinsumner:
Add an async option to PUT.

This will, as long as the slow offer state is not active (i.e. as long as there is no work backlog), respond OK before persisting the PUT.

There is no validation of the PUT at present.  The PUT will either ok/pause or crash.  This will mean that it could now ok then crash.  So caution required when implementing this.

This did speed up, and improve throughput, on counter updates - https://github.com/OpenRiak/riak_kv/issues/50#issuecomment-3346458091.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -48,7 +48,12 @@
 
 {profiles, [
     {eqc, [
-        {deps, [meck, fqc]},
+        {deps, [
+            {meck,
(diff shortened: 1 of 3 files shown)


martinsumner:
Change on hold, pending further discussion on the relative value of the throughput/performance improvement.

